PR #2, opened 2020-02-20
A skeleton version of the mock controller which was used in a demo.

wvreeven:
This is the skeleton version of the mock controller. I have adapted it from the ts_ATDome controller and used it to demo the communications protocol with EIE today. Please review and confirm the merge. If you don't have time then don't worry and I will use my admin rights again.


python/lsst/ts/dcs/__init__.py
```diff
@@ -19,4 +19,7 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
-from .version import *  # Generated by sconsUtils
+try:
+    from .version import *
+except ModuleNotFoundError:
+    __version__ = "?"
```
tribeiro: Add new line at the end of the file...

python/lsst/ts/dcs/mock_controller.py
```diff
@@ -0,0 +1,249 @@
...
+
+    async def stop(self):
+        """Stop the TCP/IP server.
+        """
+        if self._server is None:
+            return
+
+        server = self._server
+        self._server = None
+        logging.info("closing server")
```
tribeiro: Use `self.log` instead.

python/lsst/ts/dcs/mock_controller.py
```diff
@@ -0,0 +1,249 @@
...
+        server.close()
+        logging.info("done closing")
```
tribeiro: Use `self.log` instead.

python/lsst/ts/dcs/mock_controller.py
```diff
@@ -0,0 +1,249 @@
...
+                self.log.info("stopping mock controller")
+                await self.stop()
+                return
+            line = line.strip()
+            self.log.info(f"read command: {line!r}")
+            if line:
+                try:
+                    items = line.split()
+                    cmd = items[-1]
+                    if cmd not in self.dispatch_dict:
```
tribeiro: Instead of raising an exception (and capturing it out later to log/ignore) I think it would be better to respond with a failure code. This way, whoever sent the command knows it failed and why.

python/lsst/ts/dcs/mock_controller.py
```diff
@@ -0,0 +1,249 @@
...
+                        raise KeyError(f"Unsupported command {cmd}")
+                    if cmd == self.fail_command:
+                        self.fail_command = None
+                        outputs = [f"Command {cmd} failed by request"]
+                    else:
+                        has_data, func = self.dispatch_dict[cmd]
+                        desired_len = 2 if has_data else 1
+                        if len(items) != desired_len:
+                            raise RuntimeError(
```
tribeiro: Same comment as in the previous case. Instead of raising RuntimeError, I think it is better to respond with an error message.

wvreeven:
Thanks, I'll make the proposed changes tomorrow. I tried adding a unit test and failed so I will need some help there.

tribeiro:
A quick note, you may want to do `eups declare`/`setup` and run `scons` to catch pep8 violations. 

I ran it here and got quite a few...

python/lsst/ts/dcs/__init__.py
wvreeven: Thanks, done!

python/lsst/ts/dcs/mock_controller.py
wvreeven: Thanks, done!

python/lsst/ts/dcs/mock_controller.py
wvreeven: Thanks, done!

python/lsst/ts/dcs/mock_controller.py
wvreeven: Well, I copied that from the mock_controller code in ts_ATDome ;-) However, I agree that returning a failure code would be better. I'll need to discuss this with German and EIE to see how we solve this. Once we decide how I will open a new ticket to solve this and the other Exception as well.

python/lsst/ts/dcs/mock_controller.py
wvreeven: Well, I copied that from the mock_controller code in ts_ATDome ;-) However, I agree that returning a failure code would be better. I'll need to discuss this with German and EIE to see how we solve this. Once we decide how I will open a new ticket to solve this and the other Exception as well.

wvreeven:
I fixed all issues that you reported apart from the Exceptions. Can you please verify that all is OK now?

wvreeven:
I also added a unit test now. Just very basic and I'll expand it when I add more functionality to the mock controller.

PR #6, opened 2020-03-19
Tickets/dm 23842

wvreeven:


python/lsst/ts/Dome/dome_csc.py
```diff
@@ -21,47 +21,65 @@ def __init__(
...
             config_dir=config_dir,
             initial_state=initial_state,
             simulation_mode=simulation_mode,
         )
 
-    async def do_crawlAz(self, data):
+    async def do_moveAz(self, data):
         """
-        Crawl AZ
+        Move AZ
```
r-owen: Really pick point, but please put the brief description on the same line as the opening `"""`. In general we follow DM's convention for Python doc strings: https://developer.lsst.io/python/numpydoc.html (and DM basically follows numpy's convention).

python/lsst/ts/Dome/dome_csc.py
```diff
@@ -108,44 +126,88 @@ def __init__(
...
         """
-        Set Temperature
+        Internal Fans command
         """
         self.assert_enabled()
         raise salobj.ExpectedError("Not implemented")
 
+    async def inflate(self, data):
+        """
+        Internal Inflate command
```
r-owen: This is not very informative. What does "Internal" mean in this context? What does "Inflate" mean?

tests/test_dome_csc.py
```diff
@@ -13,19 +13,53 @@ def basic_make_csc(self, initial_state, config_dir, simulation_mode):
...
+                    "closeShutter",
+                    "stopShutter",
+                    "park",
+                    "setTemperature",
+                    "config",
+                    "fans",
+                    "inflate",
+                    "status",
+                )
+            )
```
r-owen: Thank you

tests/test_dome_csc.py
```diff
@@ -37,23 +71,29 @@ def basic_make_csc(self, initial_state, config_dir, simulation_mode):
...
         pass
 
-    async def test_do_setTemperature(self):
+    async def test_fans(self):
+        pass
+
+    async def test_inflate(self):
+        pass
+
+    async def test_status(self):
```
r-owen: I concerned that all these null tests will make it difficult to figure out which tests are not yet implemented (especially once you have implemented most of them). Consider replacing `pass` with `raise unittest.SkipTest("Not implemented")` or something similar.

python/lsst/ts/Dome/dome_csc.py
wvreeven: Done

python/lsst/ts/Dome/dome_csc.py
wvreeven: I added more descriptive inline comments for these "internal" functions.

tests/test_dome_csc.py
wvreeven: Done

PR #12, opened 2020-04-08
Refactored test code and added a task scheduler.

wvreeven:


wvreeven:
Thanks for the review. I converted the docstrings as much as possible to numpydoc. I hope this is satisfactory to you.

PR #13, opened 2020-04-14
Managed to get the task scheduler to run and not block the loop. The …

wvreeven:
…retrieved AMCS status of the lower level commands now gets published as telemetry.


python/lsst/ts/Dome/dome_csc.py
```diff
@@ -393,6 +400,22 @@ async def status(self):
...
+            driveTorqueCmd=self.lower_level_status["AMCS"]["driveTorqueCmd"],
+            driveCurrentActual=self.lower_level_status["AMCS"]["driveCurrentActual"],
+            driveTempActual=self.lower_level_status["AMCS"]["driveTempActual"],
+            encoderHeadRaw=self.lower_level_status["AMCS"]["encoderHeadRaw"],
+            encoderHeadCalibrated=self.lower_level_status["AMCS"][
+                "encoderHeadCalibrated"
+            ],
+            resolverRaw=self.lower_level_status["AMCS"]["resolverRaw"],
+            resolverCalibrated=self.lower_level_status["AMCS"]["resolverCalibrated"],
+        )
```
r-owen, line 418: This works, but I worry about all the repetition. A spelling or copy/paste error could be hard to spot.

It appears that `self.lower_level_status["AMCS"]` is a dict. If *all* the values in that dict are fields in the telemetry event, then this would work:
```
self.tel_domeADB_status.set_put(**self.lower_level_status["AMCS"])
```
If you are only setting a subset of the fields from `self.lower_level_status["AMCS"]` then it takes more work.
You could list the fields manually, but if the public telemetry topic fields are a subset of the fields in `self.lower_level_status["AMCS"]` then you can make the list automatically:
```
# Put this in the constructor, so you only have to do it once
# (get_vars creates a dict representation of a DDS sample)
data = self.tel_domeADB_status_fields.DataType().get_vars()
self.domeADB_status_fields = {field for field in data if not field.startswith("private_")
# If Done is indexable then delete the field "DomeID" here

# Then in this code block:
adb_status_dict = {field: self.lower_level_status["AMCS"][field] for field in self.domeADB_status_fields}
self.tel_domeADB_status.set_put(**adb_status_dict)
```
The code isn't as clear, but it eliminates a lot of repetition.

python/lsst/ts/Dome/mock_controller.py
```diff
@@ -2,7 +2,9 @@
 import logging
 import yaml
 
-logging.basicConfig(level=logging.INFO)
+logging.basicConfig(
+    format="%(asctime)s:%(levelname)s:%(name)s:%(message)s", level=logging.INFO
+)
```
r-owen, line 7: See comment above about calling basicConfig in library code. Having it here means it will be called twice, which is is not recommended.

python/lsst/ts/Dome/task_scheduler.py
```diff
@@ -1,30 +1,36 @@
...
     log.info(f"run_status_loop = {run_status_loop}")
 
     def g_tick():
-        t = loop.time()
+        t = time.time()
         count = 0
         while run_status_loop:
             count += 1
-            yield max(t + count * period - loop.time(), 0)
+            yield max(t + count * period - time.time(), 0)
```
r-owen, line 33: I'm uneasy about this class, especially when precise timing is not necessary. Compare using this class to the standard loop approach:
```
async def background_loop(self):
    while True:
        # do stuff
        await asyncio.sleep(...)
```
Personally I find the above easier to read. if you need precise timing you could write a simple functor that computes how long to sleep (a minor variant of this class), and I *think* the result would still be simpler.

python/lsst/ts/Dome/task_scheduler.py
```diff
@@ -1,30 +1,36 @@
...
         The period in (decimal) seconds at which to schedule the function.
     task: coroutine
         The function to be scheduled periodically.
     """
-    loop = asyncio.get_event_loop()
     log.info(f"run_status_loop = {run_status_loop}")
 
     def g_tick():
-        t = loop.time()
+        t = time.time()
```
r-owen, line 29: Note that `time.monotonic` is preferred when you want to write something that measures elapsed time. (Credit to John Parejko for pointing that out to me.)

python/lsst/ts/Dome/dome_csc.py
```diff
@@ -4,11 +4,14 @@
...
 import yaml
+from asyncio.streams import IncompleteReadError
 from lsst.ts import salobj
 from .mock_controller import MockDomeController
 from lsst.ts.Dome import task_scheduler
 
-logging.basicConfig(level=logging.INFO)
+logging.basicConfig(
+    format="%(asctime)s:%(levelname)s:%(name)s:%(message)s", level=logging.INFO,
+)
```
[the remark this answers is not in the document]
wvreeven: Thanks for the suggestion. You are right of course. I have removed the logging config from the library files and put them inside the unit test files and the Dome CSC start script.

python/lsst/ts/Dome/mock_controller.py
wvreeven: Thanks for the suggestion. You are right of course. I have removed the logging config from the library files and put them inside the unit test files and the Dome CSC start script.

python/lsst/ts/Dome/dome_csc.py
```diff
@@ -89,29 +93,32 @@ async def connect(self):
         )
 
         # Start polling for the status of the lower level components periodically.
-        # task_scheduler.run_status_loop = True
-        await task_scheduler.schedule_task_periodically(1, self.status)
+        task_scheduler.run_status_loop = self.run_status_loop
```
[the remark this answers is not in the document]
wvreeven: For no reason at all. I was trying to make the parameter configurable but in the end didn't need to and I forgot to remove that line and the variable. Thanks!

python/lsst/ts/Dome/dome_csc.py
```diff
@@ -89,29 +93,32 @@ async def connect(self):
...
+        asyncio.ensure_future(task_scheduler.schedule_task_periodically(1, self.status))
```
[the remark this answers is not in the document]
wvreeven: I was looking for a way to run a loop in the background and that executes a task on a fixed rate. I agree that there is no need for a very precise timing so I replaced the call to ensure_future with create_task and that's it. I honestly don't see the benefit for the make_done_future construction and I think it merely complicates things but perhaps you can explain this to me some other time again.

python/lsst/ts/Dome/task_scheduler.py
wvreeven: I was looking for a way to run a loop in the background and that executes a task on a fixed rate. I agree that there is no need for a very precise timing so I replaced most of the code with the asyncio.sleep call and that's it.

python/lsst/ts/Dome/task_scheduler.py
wvreeven: I took out the bit of the code related to this comment but still thanks for the tip. I'll try not to forget it.

python/lsst/ts/Dome/dome_csc.py
wvreeven: Thanks! I was worried about this as well. Fortunately, only one key in the dict needs to be taken out and the rest can stay. So I took a slightly different approach to simply copy the dict with one key less. I create a copy because I need to keep the original dict for internal purposes. And the code is written such that it is easy to expand the set of keys with more keys that need to be removed if necessary.

wvreeven:
Code changes pushed. Could you review once more please?

python/lsst/ts/Dome/task_scheduler.py
```diff
@@ -16,20 +15,9 @@ async def schedule_task_periodically(period, task):
...
-    def g_tick():
-        t = loop.time()
-        count = 0
-        while run_status_loop:
-            count += 1
-            yield max(t + count * period - loop.time(), 0)
-
-    g = g_tick()
-
     while run_status_loop:
```
r-owen, line 20: I strongly suggest deleting this module and ditching the global. I think the code would be much clearer as a function.
```
async def status_loop(self, period):
    while self.run_status_loop:
        self.status()
        await asyncio.sleep(period)
```

In addition, please consider canceling the task when you want to close it instead of setting a global or instance variable. The advantage is that you eliminate the latency of waiting for the sleep to finish when you want the loop to end.

That is the pattern I describe here: https://ts-salobj.lsst.io/salobj_cscs.html#telemetry-loop-example

A counter-indicator is if cancel could interrupt something that should not be interrupted -- in this case that would be something in your `status` function. But the only thing that could be interrupted is a TCP/IP read, and that should be perfectly safe. (It is rare and unwelcome to have async code that cannot safely be interrupted).

tests/test_mock_controller.py
```diff
@@ -1,10 +1,15 @@
...
 import asynctest
+import logging
 import status_assert_util as sau
 import yaml
 
 from lsst.ts import Dome
 
+logging.basicConfig(
+    format="%(asctime)s:%(levelname)s:%(name)s:%(message)s", level=logging.INFO
+)
```
r-owen, line 11: If you get tired of writing this you could provide a function that does this, preferably one that takes the level as an argument.

python/lsst/ts/Dome/dome_csc.py
```diff
@@ -4,12 +4,11 @@
 import logging
 import pathlib
 import yaml
+from asyncio.streams import IncompleteReadError
```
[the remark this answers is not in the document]
wvreeven: Done, thanks. I may need some guidance on what are considered third party libs. I suppose those that need installation via pip/conda?

python/lsst/ts/Dome/dome_csc.py
```diff
@@ -89,29 +88,32 @@ async def connect(self):
...
         # Stop polling for the status of the lower level components periodically.
         task_scheduler.run_status_loop = False
 
         writer = self.writer
         self.reader = None
         self.writer = None
+        try:
+            await self.stop_mock_ctrl()
+        except IncompleteReadError:
+            pass
```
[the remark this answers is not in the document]
wvreeven: Done, good suggestion thanks.

python/lsst/ts/Dome/dome_csc.py
```diff
@@ -159,7 +161,7 @@ async def read(self):
 
         Returns
         -------
-        data: `dictionary`
+        data: `dict`
```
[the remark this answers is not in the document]
wvreeven: Done

python/lsst/ts/Dome/dome_csc.py
```diff
@@ -394,6 +396,36 @@ async def status(self):
...
+        amcs_keys_to_remove = {"status"}
+        telemetry = self.remove_keys_from_dict(
+            self.lower_level_status["AMCS"], amcs_keys_to_remove
+        )
+        self.tel_domeADB_status.set_put(**telemetry)
+
+    # noinspection PyMethodMayBeStatic
+    def remove_keys_from_dict(self, dict_with_too_many_keys, keys_to_remove):
+        """
+        Remove the given keys (and their values) from the given dict.
```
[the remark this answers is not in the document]
wvreeven: Done both.

python/lsst/ts/Dome/task_scheduler.py
wvreeven: Yes the module really isn't necessary anymore. I removed it.

tests/test_mock_controller.py
wvreeven: Thanks, I'll remember that.

python/lsst/ts/Dome/dome_csc.py
r-owen: The new code looks great. Nice solution.

python/lsst/ts/Dome/dome_csc.py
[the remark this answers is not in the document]
r-owen: I agree the line between 3rd party packages and our packages is a bit fuzzy. I think of 3rd party packages as packages not maintained by LSST.

PR #14, opened 2020-04-23
Tickets/dm 24573

wvreeven:
I implemented a basic function to send configuration parameters to the lower level components. The function checks the configuration parameter values against the (hard coded) upper limits for those parameters. It also converts any quantity to radians if necessary since the lower level components expect radians while the configuration parameters are commanded in degrees. If a parameter value is too large or a parameter is unknown, a SAL AckError is thrown.

The function is written based on the assumption that the configuration parameters are provided in a dictionary form which directly can be passed on to the lower level components (apart from unit conversions). This assumption may not be true and will be checked at a later moment.


python/lsst/ts/Dome/configuration_limits.py
```diff
@@ -0,0 +1,34 @@
+from astropy import units as u
+import math
+
+
+configuration_limits = {
+    "AMCS": {
+        "jmax": {
+            "description": "AMCS Maximum Jerk",
+            "upper_limit": (3.0 * math.pi / 180.0) * u.rad / u.s ** 3,
```
r-owen, line 9: Looking at `config_lls` I don't think you are getting any value out of these being astropy quantities. It seems to me that what you want to do is easily convert between degrees and radians, while ignoring the time component, and astropy just doesn't seem to support that.

Please consider just using radians or degrees here. Personally I suggest degrees, for ease of use, but pick one and stick to it. You can use one code comment to document your convention (that seems less bulky than a separate units string per entry, which would be hard to enforce).

Also please consider a constant such as `RADIANS_PER_DEGREES`. It may be a few more characters but I think `3.0 * RADIANS_PER_DEGREES` is easier to read and the intent is clearer.

python/lsst/ts/Dome/dome_csc.py
```diff
@@ -361,7 +362,45 @@ async def config(self, _data):
             The contents of this parameter will be defined soon.
         """
```
r-owen, line 363: "will be defined soon"? Can you define it now?

python/lsst/ts/Dome/dome_csc.py
```diff
@@ -361,7 +362,45 @@ async def config(self, _data):
...
         self.assert_enabled()
```
r-owen, line 364: Do you need to check enabled for non-SAL commands?

python/lsst/ts/Dome/configuration_limits.py
r-owen: Please consider replacing the dicts with a small struct-like custom class. Advantages:
* You need to check that fields are present; they are sure to be present.
* You can document the constructor parameters -- including the units.
* There is no need for the "description" and "upper_limit" fields -- the parameter documentation does all that for you.
* You can add a method to validate a dict of proposed values (e.g. in config_lls).
* You can even add a name field (that you could set to "AMCS" or "LWSCS"). I doubt you would find it helpful, but it might come in handy for error messages.

Then all this code becomes:
```
configuration_limits = dict(
    AMCS=ConfigurationLimit(jmax=3.0, amax=0.5, vmax=1.5),
    LWSCS=ConfigurationLimit(jmax=3.5, amax=0.875, vmax=1.75)
)
```
with the class defined somewhere. By the time you define the class there may be more net code, but I think the result will be significantly easier to read and use. Something along the line of:
```
class ConfigurationLimit:
    """...docs including the crucial Parameters section..."""
    def __init__(self, jmax, amax, vmax):
        self.jmax = jmax
        self.amax = amax
        self.vmax = vmax

   def check_config_data(self, data):
        if data["vmax"] > self.vmax * RADIANS_PER_DEGREE:
            raise ValueError(f"...")
```

python/lsst/ts/Dome/mock_controller.py
```diff
@@ -243,6 +248,9 @@ async def stop_el(self):
         self.log.info("Received command 'stopEl'")
         self.el_motion = "Stopped"
 
+    async def config(self, **kwargs):
+        self.log.info(f"Received command 'moveAz' with arguments {kwargs}")
```
r-owen, line 252: Please update the string. If you want to print such a message for every command method then please consider logging the message in the code that calls the method -- one message would handle all commands.

wvreeven: This is a copy and paste error and I will fix it.

python/lsst/ts/Dome/dome_csc.py
wvreeven: Good question and it is one that I have been struggling with. Currently the mock_controller only gets started if the DCS state is disabled or enable. There doesn't seem to be an "assert_disabled_or_enabled()" function in SALobj.

python/lsst/ts/Dome/dome_csc.py
wvreeven: To be honest, I haven't got a clue how configuration parameters will be passed on from TCS to DCS so the answer to your question is "no".

python/lsst/ts/Dome/dome_csc.py
r-owen: Correct. There is an `enabled_or_disabled` property and it's easy enough to explicitly test for the desired states. (Either way, it's best to raise `salobj.ExpectedError` to avoid a pointless traceback.)

python/lsst/ts/Dome/configuration_limits.py
wvreeven: I overhauled this code completely based on the Slack chats we had today. Many thanks for all your inputs!

python/lsst/ts/Dome/dome_csc.py
wvreeven: I have removed this assert for now, as well in other methods that are not being called by SAL directly. It is clear to me that a discussion is needed about configuring this CSC and until then I will leave the method in place to not have wasted time.

wvreeven:
New commit pushed. Can you please review again?

python/lsst/ts/Dome/dome_csc.py
```diff
@@ -183,8 +188,8 @@ async def do_moveAz(self, _data):
 
         Parameters
         ----------
-        _data : `A SALOBJ data object`
-            Contains the data as defined in the SAL XML file.
+        _data : `A SALOBJ configuration_parameters object`
```
r-owen, line 191: I think this is a search/replace error. Also...why the leading underscore on the data argument?

By the way: the back ticks tell sphinx to try to make a link, so they are only appropriate if the type is an actual type.

wvreeven: Thanks. PyCharm must have messed up due to some refactoring that I did.

The underscore prevents PyCharm from complaining when the parameter is not used. In cases where it IS used I could remove them. But perhaps for uniformity of the API I can leave them in place everywhere. I do prefer to keep the underscore since it is one warning less and I prefer to have my IDE without warnings.

wvreeven:
I fixed the doc string problem. Can you have another look please?

wvreeven:
Pushed a new version of dome_csc.py with all parameters _data renamed to data.

python/lsst/ts/Dome/llc_configuration_limits/abstract_llc_configuration_limits.py
```diff
@@ -0,0 +1,14 @@
+from abc import ABC, abstractmethod
+
+
+class AbstractLlcConfigurationLimits(ABC):
+    """An abstract base class for holding the configuration limits for the lower level components. The
+    class holds common methods and constant values.
```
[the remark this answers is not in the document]
wvreeven: This always will be limits. If this changes in the future then I can consider renaming the class.

python/lsst/ts/Dome/llc_configuration_limits/amcs_configuration_limits.py
```diff
@@ -0,0 +1,85 @@
+from .abstract_llc_configuration_limits import AbstractLlcConfigurationLimits
+import math
+
+DEGREES_TO_RADIANS = math.pi / 180.0
+
+
+class AmcsConfigurationLimits(AbstractLlcConfigurationLimits):
+    """This class holds the limits of the configuration values for the AMCS lower level component. It will
+    validate any set of configuration parameters againt these limits. It will also convert any
+    configuration parameter values expressed in deg/s^n to rad/s^n (with n = 1, 2 or 3).
```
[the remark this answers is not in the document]
wvreeven: Strictly speaking they are not parameters since they cannot be passed as arguments to the constructor. So I added a similar doc string since I agree that a description there is very useful.

python/lsst/ts/Dome/llc_configuration_limits/amcs_configuration_limits.py
```diff
@@ -0,0 +1,85 @@
...
+        converted_configuration_parameters = {}
+
+        # Validate the provided value for max jerk (in deg/s^3) against the limit (in deg/s^3).
+        if configuration_parameters["jmax"] <= self.jmax:
+            # If the value is smaller than or equal to the limit, convert to rad/s^3.
+            converted_configuration_parameters["jmax"] = (
+                configuration_parameters["jmax"] * DEGREES_TO_RADIANS
+            )
+            # Remove the key so we can check later if any unknown configuration parameters remain.
+            del configuration_parameters["jmax"]
```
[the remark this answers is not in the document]
wvreeven: That's a good suggestion and will work for now. However, the list of defined parameters may get large and I may switch to a dict then to avoid duplicating the list of expected keys.

schema/Dome.yaml
```diff
@@ -20,7 +20,7 @@ properties:
     exclusiveMinimum: 0
     default: 10
   read_timeout:
-    description: Time limit for reading data from the TCP/IP interface (sec)
+    description: Time limit for reading configuration_parameters from the TCP/IP interface (sec)
```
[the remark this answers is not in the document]
wvreeven: Thanks for catching this.

tests/status_assert_util.py
```diff
@@ -2,7 +2,7 @@
 
 
 def assertReply(component, lower_level_status, **kwargs):
-    """Asserts that the values of the component parameter data are as expected.
+    """Asserts that the values of the component parameter configuration_parameters are as expected.
```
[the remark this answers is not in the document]
wvreeven: Thanks for catching this.

tests/status_assert_util.py
```diff
@@ -20,7 +20,7 @@ def assertReply(component, lower_level_status, **kwargs):
 
 
 def assertTBD(component, lower_level_status):
-    """Asserts that the values of the component parameter data are "TBD"
+    """Asserts that the values of the component parameter configuration_parameters are "TBD"
```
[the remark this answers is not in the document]
wvreeven: Thanks for catching this.

tests/test_dome_csc.py
```diff
@@ -14,7 +14,7 @@
 
 
 class CscTestCase(salobj.BaseCscTestCase, asynctest.TestCase):
-    def basic_make_csc(self, initial_state, config_dir, simulation_mode):
+    def basic_make_csc(self, initial_state, config_dir, simulation_mode, **kwargs):
```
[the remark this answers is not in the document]
wvreeven: I added the **kwargs argument because PyCharm complains about

"Signature of method 'CscTestCase.basic_make_csc()' does not match signature of base method in class 'BaseCscTestCase'"

The BaseCscTestCase make_csc function indeed contains an additional **kwargs parameter.

tests/test_mock_controller.py
```diff
@@ -31,7 +31,7 @@ async def read(self):
 
         Returns
         -------
-        data : `dictionary`
+        configuration_parameters : `dictionary`
```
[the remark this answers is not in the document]
wvreeven: Thanks, I modified dictionary to dict and string to str.

python/lsst/ts/Dome/llc_configuration_limits/amcs_configuration_limits.py
```diff
@@ -0,0 +1,85 @@
...
+    configuration parameter values expressed in deg/s^n to rad/s^n (with n = 1, 2 or 3).
+    """
+
+    def __init__(self):
+        self.jmax = 3.0  # Maximum jerk in deg/s^3
+        self.amax = 0.75  # Maximum acceleration in deg/s^2
+        self.vmax = 1.5  # Maximum velocity in deg/s
+
+    def validate_and_convert_from_degrees_to_radians(self, configuration_parameters):
+        """Validates the data are against the configuration limits of the lower level component. If
```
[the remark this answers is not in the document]
wvreeven: Thanks, I was in doubt about this so it is clear to me now.

python/lsst/ts/Dome/mock_controller.py
```diff
@@ -99,11 +100,15 @@ async def cmd_loop(self, reader, writer):
...
-            self.log.info(f"Read command line: {line!r}")
             timeout = 20
+            print_ok = True
+            line = None
+            try:
+                line = await reader.readuntil(b"\r\n")
+                line = line.decode().strip()
+                self.log.info(f"Read command line: {line!r}")
+            except asyncio.IncompleteReadError:
+                pass
```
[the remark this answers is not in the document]
wvreeven: That is a good suggestion. However, for debugging purposes I find it very useful to have a log line at the start of each function so I can follow the process flow more easily.

tests/test_dome_csc.py
```diff
@@ -125,7 +125,37 @@ async def test_do_setTemperature(self):
...
+            # The param AMCS smax doesn't exist.
+            config = {
+                "AMCS": {"jmax": 1.0, "amax": 0.5, "vmax": 1.0, "smax": 1.0},
+                "LWSCS": {"jmax": 1.0, "amax": 0.5, "vmax": 1.0},
+            }
+            try:
+                await self.csc.config_llcs(config)
+                self.fail("Expected a KeyError.")
+            except KeyError:
+                pass
```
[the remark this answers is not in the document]
wvreeven: Thanks, I checked the meeting minutes and we indeed agreed to always send the complete set of parameters. I will add the check.

As for the mock controller: currently it simply accepts whatever is sent. I hope to not to have to spend much more time on the mock controller and that EIE will come up with a version of their software supporting a simulator mode.

python/lsst/ts/Dome/llc_configuration_limits/lwscs_configuration_limits.py
```diff
@@ -0,0 +1,85 @@
+from .abstract_llc_configuration_limits import AbstractLlcConfigurationLimits
+import math
+
+DEGREES_TO_RADIANS = math.pi / 180.0
+
+
+class LwscsConfigurationLimits(AbstractLlcConfigurationLimits):
```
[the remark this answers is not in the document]
wvreeven: I opted to introduce a class that validates the common parameters shared between AMCS and LWSCS as a sub-class of the abstract validation class. The AMCS and LWSCS classes are now sub-classes of that new sub-class.

PR #17, opened 2020-05-12
Made sure that the Mock Controller uses radians everywhere. As a cons…

wvreeven:
…equence the DCS now performs conversions from degrees to radians and back where necessary.

### DIFF
--- a/python/lsst/ts/Dome/dome_csc.py
+++ b/python/lsst/ts/Dome/dome_csc.py
@@ -3,6 +3,8 @@ __all__ = ["DomeCsc"]
 import asyncio
 import logging
 import pathlib
+import math
+import numpy as np
 import yaml
 
 from .llc_configuration_limits import AmcsLimits, LwscsLimits
@@ -12,6 +14,10 @@ from .mock_controller import MockDomeController
 
 _LOCAL_HOST = "127.0.0.1"
 _TIMEOUT = 20  # timeout in s to be used by this module
+_DEGREES_TO_RADIANS = math.pi / 180.0
+_RADIANS_TO_DEGREES = 180.0 / math.pi
+_KEYS_TO_REMOVE = {"status"}
+_KEYS_IN_RADIANS = {"positionError", "positionActual", "positionCmd"}
 
 
 class DomeCsc(salobj.ConfigurableCsc):
@@ -193,8 +199,7 @@ class DomeCsc(salobj.ConfigurableCsc):
             Contains the data as defined in the SAL XML file.
         """
         self.assert_enabled()
-        # TODO Make sure that radians are used because that is what the real LLCs will use as well. DM-24789
-        cmd = {"moveAz": {"azimuth": data.azimuth}}
+        cmd = {"moveAz": {"azimuth": data.azimuth * _DEGREES_TO_RADIANS}}
         self.log.info(f"Moving Dome to azimuth {data.azimuth}")
         await self.write_then_read_reply(cmd)
 
@@ -207,8 +212,7 @@ class DomeCsc(salobj.ConfigurableCsc):
             Contains the data as defined in the SAL XML file.
         """
         self.assert_enabled()
-        # TODO Make sure that radians are used because that is what the real LLCs will use as well. DM-24789
-        cmd = {"moveEl": {"elevation": data.elevation}}
+        cmd = {"moveEl": {"elevation": data.elevation * _DEGREES_TO_RADIANS}}
         self.log.info(f"Moving LWS to elevation {data.elevation}")
         await self.write_then_read_reply(cmd)
 
@@ -257,8 +261,12 @@ class DomeCsc(salobj.ConfigurableCsc):
             Contains the data as defined in the SAL XML file.
         """
         self.assert_enabled()
-        # TODO Make sure that radians are used because that is what the real LLCs will use as well. DM-24789
-        cmd = {"crawlAz": {"dirMotion": data.dirMotion, "azRate": data.azRate}}
+        cmd = {
+            "crawlAz": {
+                "dirMotion": data.dirMotion,
+                "azRate": data.azRate * _DEGREES_TO_RADIANS,
+            }
+        }
         await self.write_then_read_reply(cmd)
 
     async def do_crawlEl(self, data):
@@ -270,8 +278,12 @@ class DomeCsc(salobj.ConfigurableCsc):
             Contains the data as defined in the SAL XML file.
         """
         self.assert_enabled()
-        # TODO Make sure that radians are used because that is what the real LLCs will use as well. DM-24789
-        cmd = {"crawlAz": {"dirMotion": data.dirMotion, "elRate": data.elRate}}
+        cmd = {
+            "crawlAz": {
+                "dirMotion": data.dirMotion,
+                "elRate": data.elRate * _DEGREES_TO_RADIANS,
+            }
+        }
         await self.write_then_read_reply(cmd)
 
     async def do_setLouver(self, data):
@@ -283,8 +295,12 @@ class DomeCsc(salobj.ConfigurableCsc):
             Contains the data as defined in the SAL XML file.
         """
         self.assert_enabled()
-        # TODO Make sure that radians are used because that is what the real LLCs will use as well. DM-24789
-        cmd = {"setLouver": {"id": data.id, "position": data.position}}
+        cmd = {
+            "setLouver": {
+                "id": data.id,
+                "position": data.position * _DEGREES_TO_RADIANS,
+            }
+        }
         await self.write_then_read_reply(cmd)
 
     async def do_closeLouvers(self, data):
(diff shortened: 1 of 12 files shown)


python/lsst/ts/Dome/dome_csc.py
```diff
@@ -12,6 +14,10 @@
 
 _LOCAL_HOST = "127.0.0.1"
 _TIMEOUT = 20  # timeout in s to be used by this module
+_DEGREES_TO_RADIANS = math.pi / 180.0
```
pkubanek, line 17: numpy provides numpy.degrees and numpy.radians. Please use those and you wan't need the constants:

```python
In [1]: import numpy                                                                                     

In [2]: numpy.degrees(180)                                                                               
Out[2]: 10313.240312354817

In [3]: numpy.radians(180)                                                                               
Out[3]: 3.141592653589793

In [4]: numpy.radians(numpy.pi)                                                                          
Out[4]: 0.05483113556160755
```

python/lsst/ts/Dome/dome_csc.py
```diff
@@ -207,8 +212,7 @@ async def do_moveEl(self, data):
             Contains the data as defined in the SAL XML file.
         """
         self.assert_enabled()
-        # TODO Make sure that radians are used because that is what the real LLCs will use as well. DM-24789
-        cmd = {"moveEl": {"elevation": data.elevation}}
+        cmd = {"moveEl": {"elevation": data.elevation * _DEGREES_TO_RADIANS}}
```
pkubanek, line 215: so this will be numpy (or np if you use standard import numpy as np) .radians(data.elevation)..etc..

tests/test_mock_controller.py
```diff
@@ -85,10 +88,10 @@ async def test_moveAz(self):
             amcs_status["status"], Dome.LlcStatus.MOVING.value,
         )
         self.assertGreaterEqual(
-            amcs_status["positionActual"], 0.5,
+            amcs_status["positionActual"], 0.5 * _DEGREES_TO_RADIANS,
```
pkubanek, line 91: consider import numpy as np, and then

```python
amcs_status["positionActual"], np.radians(0.5),
```

python/lsst/ts/Dome/dome_csc.py
r-owen: I gently disagree. For scalars it's better to use the above or `math.degrees` and `math.radians` because numpy is slow for scalars. Personally I have a slight aversion to the functions because I feel their meaning is less obvious than using a named constant, so I usually use a named constant. But either is fine.

python/lsst/ts/Dome/dome_csc.py
```diff
@@ -3,6 +3,8 @@
 import asyncio
 import logging
 import pathlib
+import math
+import numpy as np
```
r-owen, line 7: Please insert a blank line before the 3rd party imports

python/lsst/ts/Dome/dome_csc.py
```diff
@@ -506,10 +519,33 @@ def remove_keys_from_dict(self, dict_with_too_many_keys, keys_to_remove):
...
+    # noinspection PyMethodMayBeStatic
+    def convert_telemetry_radians_to_degrees(self, telemetry_in_radians):
+        telemetry_in_degrees = {}
+        for key in telemetry_in_radians.keys():
+            if key in _KEYS_IN_RADIANS:
+                telemetry_in_degrees[key] = (
+                    telemetry_in_radians[key] * _RADIANS_TO_DEGREES
+                )
+            else:
+                telemetry_in_degrees[key] = telemetry_in_radians[key]
```
r-owen, line 535: Nice way to handle this. The "else" caught me off guard at first (setting degrees to radians?), but I have no idea how to make it less startling, except perhaps a code comment such as "not an angle".

python/lsst/ts/Dome/dome_csc.py
```diff
@@ -506,10 +519,33 @@ def remove_keys_from_dict(self, dict_with_too_many_keys, keys_to_remove):
...
+    # noinspection PyMethodMayBeStatic
+    def convert_telemetry_list_radians_to_degrees(self, telemetry_in_radians):
+        telemetry_in_degrees = {}
+        for key in telemetry_in_radians.keys():
+            if key in _KEYS_IN_RADIANS:
+                arr = np.array(telemetry_in_radians[key])
+                telemetry_in_degrees[key] = (arr * _RADIANS_TO_DEGREES).tolist()
+            else:
+                telemetry_in_degrees[key] = telemetry_in_radians[key]
+        return telemetry_in_degrees
```
r-owen, line 547: @pkubanek 's suggestion to use `numpy.degrees` would work well here, since I think you can use the same code for both lists and scalars

That only works if you are willing to have the values be numpy arrays instead of lists. But if that works then I would argue it's actually better to have the values be numpy arrays.

python/lsst/ts/Dome/dome_csc.py
r-owen: If you are only using this for scalars then I suggest using the math functions or named constants instead. But if you have sequences, or a mix of sequences and scalars then I agree that numpy is the way to go.

python/lsst/ts/Dome/llc_configuration_limits/amcs_limits.py
```diff
@@ -1,5 +1,9 @@
+import math
+
 from .common_amcs_and_lwscs_limits import CommonAmcsAndLwscsLimits
 
+_DEGREES_TO_RADIANS = math.pi / 180.0
```
r-owen, line 5: If you do go with a constant instead of the functions @pkubanek suggested then I suggest you only define this in one module, as a public constant, and import it everywhere.

python/lsst/ts/Dome/mock_llc_statuses/amcs_status.py
```diff
@@ -7,7 +7,7 @@
 from ..azcs_motion_direction import AzcsMotionDirection as motion_dir
 from ..llc_status import LlcStatus
 
-NUM_MOTORS = 5
+_NUM_MOTORS = 5
```
r-owen, line 10: For the sake of unit tests I wonder if you might want this information to be public. But if so, you'd probably want to qualify it somehow (number of what kind of motors?).

You might consider making it a class variable instead of a module constant:
```
class AmcsStatus(...):
    num_motors = 5

    def __init__(self, period):
        ...
```

python/lsst/ts/Dome/dome_csc.py
pkubanek: I did some timeing, using timeit in ipython. numpy.radians is ~10x slower than math.radians. math.radians is marginally (~2x with big dev) slower than raw multiplications. raw multiplications (1000x in a for loop) is ~36us, math.radians is ~73us, numpy.radians is ~770us.

pkubanek:
```python
In [40]: def rd_np(): 
    ...:     for x in range(0,1000): 
    ...:         d=np.radians(180) 
    ...:                                                                                                 

In [41]: timeit rd_np()                                                                                  
759 µs ± 3.95 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

In [42]: def rd_math(): 
    ...:     for x in range(0,1000): 
    ...:         d=math.radians(180) 
    ...:          
    ...:                                                                                                 

In [43]: timeit rd_math()                                                                                
73.8 µs ± 115 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)

In [46]: _a = math.pi / 180.0                                                                            

In [47]: _a                                                                                              
Out[47]: 0.017453292519943295

In [48]: def rd_raw(): 
    ...:     for x in range(0,1000): 
    ...:         d=180*_a 
    ...:                                                                                                 

In [49]: timeit rd_raw()                                                                                 
46.8 µs ± 75.1 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```

r-owen: I prefer `math` to `numpy` when there are only scalars because `math` is built in, and I feel it makes the intent of the code clearer (since `math` only works with scalars). Also numpy scalars can be a bit surprising at times, though they do a good job of imitating standard Python numeric types, so it is rarely an issue.

I heartily agree that `numpy` is the way to go if the code uses arrays or a mix of arrays and scalars. It is rare that speed of numpy with scalars is an issue, though in time critical applications it pays to be careful. I recently ran into a serious performance issue that I traced to slow performance of `astropy.time` with scalars.

python/lsst/ts/Dome/dome_csc.py
wvreeven: Thanks, you made this comment before. I am still getting used to this. Done.

python/lsst/ts/Dome/dome_csc.py
wvreeven: Done.

python/lsst/ts/Dome/dome_csc.py
wvreeven: Unfortunately I cannot use numpy arrays everywhere mostly because yaml doesn't handle them well and lists of values are needed for the communication with the lower level components. I will modify the code to use numpy for the list conversion though since it is much simpler and clearer. Also, I got rid of the constants and am using the math.radians and math.degrees functions now everywhere since I think that is clearer and more concise as well.

python/lsst/ts/Dome/llc_configuration_limits/amcs_limits.py
wvreeven: I removed the constants everywhere and replaced them with the math functions, except in one case where using the numpy function produces clearer code.

python/lsst/ts/Dome/dome_csc.py
wvreeven: I resolved to using meth.radians and math.degrees in almost all cases, except where a list was involved where I am using numpy instead.

tests/test_mock_controller.py
wvreeven: I resolved to using math.radians and math.degrees in almost all cases, except where a list was involved where I am using numpy instead.

python/lsst/ts/Dome/mock_llc_statuses/amcs_status.py
wvreeven: Thanks, I'll take that under consideration.

PR #20, opened 2020-05-27
Replaced the timed status loop with a method that determines the time…

wvreeven:
… difference since the last invocation. Introduced a Mock solution to avoid calling of the time difference method by the mock controller in unit tests.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -45,9 +45,9 @@ pipeline {
         }
         stage("Checkout xml") {
             steps {
-                latest_full_tag = sh(returnStdout:  true, script: "git ls-remote --tags https://github.com/lsst-ts/ts_xml | grep refs\\/tags\\/v[0-9] | tail -n 1").trim()
-                latest_tag = latest_full_tag.replaceFirst(/.*refs\//, "").replaceFirst(/\^.*/, "")
                 script {
+                    latest_full_tag = sh(returnStdout:  true, script: "git ls-remote --tags https://github.com/lsst-ts/ts_xml | grep refs\\/tags\\/v[0-9] | tail -n 1").trim()
+                    latest_tag = latest_full_tag.replaceFirst(/.*refs\//, "").replaceFirst(/\^.*/, "")
                     sh """
                     docker exec -u saluser \${container_name} sh -c \"source ~/.setup.sh && cd /home/saluser/repos/ts_xml && /home/saluser/.checkout_repo.sh tags/\${latest_tag} && git pull\"
                     """
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -45,7 +45,7 @@ pipeline {
         }
         stage("Checkout xml") {
             steps {
-                latest_tag = sh(returnStdout:  true, script: "git ls-remote --tags https://github.com/lsst-ts/ts_xml | grep refs\\/tags\\/v[0-9] | tail -n 1 | sed 's/.*\\///' | sed 's/\\^.*//'").trim()
+                latest_tag = sh(returnStdout:  true, script: "git ls-remote --tags https://github.com/lsst-ts/ts_xml | grep refs\\/tags\\/v[0-9] | tail -n 1 | sed 's/\.*\\///' | sed 's/\\^\.*//'").trim()
                 script {
                     sh """
                     docker exec -u saluser \${container_name} sh -c \"source ~/.setup.sh && cd /home/saluser/repos/ts_xml && /home/saluser/.checkout_repo.sh tags/\${latest_tag} && git pull\"
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -45,7 +45,7 @@ pipeline {
         }
         stage("Checkout xml") {
             steps {
-                latest_tag = sh(returnStdout:  true, script: "git ls-remote --tags https://github.com/lsst-ts/ts_xml | grep refs\/tags\/v[0-9] | tail -n 1 | sed 's/.*\///' | sed 's/\^.*//'").trim()
+                latest_tag = sh(returnStdout:  true, script: "git ls-remote --tags https://github.com/lsst-ts/ts_xml | grep refs\\/tags\\/v[0-9] | tail -n 1 | sed 's/.*\\///' | sed 's/\\^.*//'").trim()
                 script {
                     sh """
                     docker exec -u saluser \${container_name} sh -c \"source ~/.setup.sh && cd /home/saluser/repos/ts_xml && /home/saluser/.checkout_repo.sh tags/\${latest_tag} && git pull\"
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -45,7 +45,8 @@ pipeline {
         }
         stage("Checkout xml") {
             steps {
-                latest_tag = sh(returnStdout:  true, script: "git ls-remote --tags https://github.com/lsst-ts/ts_xml | grep refs\\/tags\\/v[0-9] | tail -n 1 | sed 's/.*\\///' | sed 's/\\^.*//'").trim()
+                latest_full_tag = sh(returnStdout:  true, script: "git ls-remote --tags https://github.com/lsst-ts/ts_xml | grep refs\\/tags\\/v[0-9] | tail -n 1").trim()
+                latest_tag = latest_full_tag.replaceFirst(/.*refs\//, "").replaceFirst(/\^.*/, "")
                 script {
                     sh """
                     docker exec -u saluser \${container_name} sh -c \"source ~/.setup.sh && cd /home/saluser/repos/ts_xml && /home/saluser/.checkout_repo.sh tags/\${latest_tag} && git pull\"
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -46,10 +46,9 @@ pipeline {
         stage("Checkout xml") {
             steps {
                 script {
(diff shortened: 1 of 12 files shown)


python/lsst/ts/Dome/dome_csc.py
```diff
@@ -178,8 +178,6 @@ async def write_then_read_reply(self, cmd):
             A dict of the form {"reply": {"param1": value1, "param2": value2}} where "reply" can for
             instance be "OK" or "ERROR".
          """
-        # TODO Create a function that takes the command and its parameters and returns a yaml
-        #  representation of it so all do_XXX functions below can be simplified. DM-24776
```
r-owen, line 182: What I had in mind is to change this method to take the command and parameters, so the user doesn't have to know the details of the form sent to the low-level controller:
```
def write_then_read_reply(self, cmd, **params):
    st = yaml.safe_dump({cmd: **params}, default_flow_style=None)
    self.writer.write(st.encode() + b"\r\n")
```
Here are two examples of usage:
```
await self.write_then_read_reply("crawlEl", dirMotion=data.dirMotion, elRate=math.radians(data.elRate))
await self.write_then_read_reply("stopEl")
```

python/lsst/ts/Dome/mock_controller.py
```diff
@@ -66,17 +68,18 @@ def __init__(
         # Name of a command to report as failed once, the next time it is seen,
         # or None if no failures. Used to test CSC handling of failed commands.
         self.fail_command = None
-        self.status_task = None
-        self.period = 0.1
         # Timeouts used by this class and by its unit test
         self.long_timeout = 20
         self.short_timeout = 2
+        # Time keeping variables
+        self.previous_time_sec = 0
```
r-owen, line 75: Please consider this slightly different implementation:
- Specify the time as an argument to determine_statuses, in TAI unix seconds (NOT a duration).
- For mock actuators: record the starting time of each move (and leave it alone until the next move)

Advantages:
- You know at exactly what time status was determined and can provide that time in the status message(s). This is certainly best practice for a real controller.
- It makes unit tests easy to write since you can specify any time you like, and you know the starting time for each actuator so you know what status to expect.

With such a design I don't think it is necessary to record a time here.

python/lsst/ts/Dome/dome_csc.py
wvreeven: Thanks for the explanation. We will replace YAML with JSON soon so I will take care of this then. See

https://jira.lsstcorp.org/browse/DM-25130

wvreeven: Done

python/lsst/ts/Dome/mock_controller.py
wvreeven: Done

wvreeven:
Changes pushed to GitHub. Can you please have another look?

python/lsst/ts/Dome/dome_csc.py
r-owen: Thanks. I'm glad you found it useful.

python/lsst/ts/Dome/mock_controller.py
```diff
@@ -240,6 +224,22 @@ async def status(self):
...
+        """
+        self.log.debug(f"self.current_tai = {self.current_tai}")
+        await self.amcs.determine_status(self.current_tai)
+        await self.apscs.determine_status(self.current_tai)
+        await self.lcs.determine_status(self.current_tai)
+        await self.lwscs.determine_status(self.current_tai)
+        await self.moncs.determine_status(self.current_tai)
+        await self.thcs.determine_status(self.current_tai)
+
+    async def determine_current_tai(self):
```
r-owen, line 238: Please consider deleting this method and just calling determine_statuses with the desired (presumably current) time.

If you want to record the time at which status was most recently computed (I have no idea if it's useful), you can do that as the first line of determine_statuses.

python/lsst/ts/Dome/mock_llc_statuses/amcs_status.py
```diff
@@ -50,24 +45,30 @@ def __init__(self, period):
         self.resolver_raw = np.zeros(_NUM_MOTORS, dtype=float)
         self.resolver_calibrated = np.zeros(_NUM_MOTORS, dtype=float)
 
-    async def determine_status(self):
+    async def determine_status(self, current_tai):
```
r-owen, line 48: Please document the new argument in the doc string.

python/lsst/ts/Dome/mock_llc_statuses/lwscs_status.py
```diff
@@ -102,7 +100,7 @@ async def moveEl(self, elevation):
             self.motion_direction = motion_dir.DOWN.value
 
     async def crawlEl(self, direction, velocity):
-        """Mock crawling of the light and wind screen in the given direction at the given velocity.
+        """Crawl the light and wind screen in the given direction at the given velocity.
```
r-owen, line 103: I'm curious: are we going to use this command? I thought you or German said no. If you are *sure* we won't want it then please consider taking it out.

tests/test_mock_controller.py
```diff
@@ -2,19 +2,31 @@
...
 _NUM_THERMO_SENSORS = 16
 
 
+# MagicMock doesn't support async coroutines and AsyncMock is not available in Python 3.7 but only in 3.8
+# This is a way to provide a future that can be returned by a mock method and that can awaited in the
+# calling code.
+def async_return(result):
+    f = asyncio.Future()
+    f.set_result(result)
+    return f
```
r-owen, line 27: Note that we use the `asynctest` package which has its own asyncio-friendly MagicMock class. It'll be interesting to see if we keep using it for the long run. I was disappointed when I read about the asyncio support in Python 3.8 unittest -- asynctest seems significantly easier to use. But we'll have to move away if it is abandoned.

tests/test_mock_controller.py
```diff
@@ -531,8 +639,14 @@ async def test_setTemperature(self):
...
         self.assertEqual(self.data["OK"]["Timeout"], self.mock_ctrl.long_timeout)
+
+        # Set the TAI time in the mock controller for easier control
+        self.mock_ctrl.current_tai = salobj.current_tai()
+        # Set the mock device statuses TAI time to the mock controller time for easier control
+        self.mock_ctrl.thcs.command_time_tai = self.mock_ctrl.current_tai
         # Give some time to the mock device to set the temperature.
-        await asyncio.sleep(0.2)
+        self.mock_ctrl.current_tai = self.mock_ctrl.current_tai + 0.2
+
```
r-owen, line 649: I hope you are testing the positions of your mock actuators somewhere. (One of the big advantages of the new way of managing time is you can easily specify a given time and know what position to expect). I'm not seeing it in the diffs above.

Jenkinsfile
```diff
@@ -47,7 +48,7 @@ pipeline {
             steps {
                 script {
                     sh """
-                    docker exec -u saluser \${container_name} sh -c \"source ~/.setup.sh && cd /home/saluser/repos/ts_xml && /home/saluser/.checkout_repo.sh \${work_branches} && git pull\"
```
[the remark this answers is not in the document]
wvreeven: I reverted al changes to Jenkinsfile.

python/lsst/ts/Dome/dome_csc.py
```diff
@@ -163,24 +163,29 @@ async def handle_summary_state(self):
         else:
             await self.disconnect()
 
-    async def write_then_read_reply(self, cmd):
+    async def write_then_read_reply(self, cmd, **params):
         """Write the cmd string appended with a newline character and then read the reply to the command.
 
+        The command and parameters will first be transformed to a dict of the form::
```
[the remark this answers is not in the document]
wvreeven: Good points. Done.

python/lsst/ts/Dome/mock_controller.py
wvreeven: The reason that I introduced this method is to be able to mock in in the unit tests so I can control the time more exactly. If the determine_statuses method determines the status then it will be much harder to mock the time. So I prefer to keep it.

python/lsst/ts/Dome/mock_llc_statuses/lwscs_status.py
wvreeven: It is not sure yet so I will leave it for now. As soon as it is clear that we will not need it, I will remove it. It then would require a change in ts_xml as well.

python/lsst/ts/Dome/mock_llc_statuses/amcs_status.py
wvreeven: It is documented in the BaseMockStatus class. As far as I know, there is no way to refer to the doc string of the super class but if there is I'd like to know and I will add it to the six sub classes.

tests/test_mock_controller.py
wvreeven: You don't see that in the differences because this is part of the test_mock_controller test and it was already added there when the timed loop was added. There are no separate test files for the actuators yet and I will consider adding them.

tests/test_mock_controller.py
wvreeven: It turns out that MagicMock in asynctest also cannot be used to mock methods that need to be awaited. But using CoroutineMock, as suggested by the docs of asynctest, works. Thanks for the tip!

PR #23, opened 2020-06-01
Removed the direction arguments from crawlAz and crawlEl.

wvreeven:


python/lsst/ts/Dome/mock_controller.py
```diff
@@ -247,13 +247,16 @@ async def move_az(self, **kwargs):
         ----------
         kwargs: `dict`
             A dictionary with arguments to the function call. It should contain the key "azimuth" with a
-            float value.
+            float value and the key "azRate" with a signed float value where positive means towards
```
r-owen, line 250: Please consider deleting "signed" from "signed float value" here and below.

r-owen: Can `azimuth` be signed? If it has a required range this seems like a good place to state it. (I assume you will match the vendor's code, but I have no idea what it does).

python/lsst/ts/Dome/mock_llc_statuses/amcs_status.py
```diff
@@ -26,7 +25,7 @@ def __init__(self):
         self.vmax = self.amcs_limits.vmax
         # variables helping with the state of the mock AZ motion
         self.motion_velocity = self.vmax
-        self.motion_direction = motion_dir.CW.value
+        self.crawl_velocity = 0
```
r-owen, line 28: Please add code comments explaining what `motion_velocity` and `crawl_velocity` are. I am guessing that motion_velocity is the speed during the move and crawl_velocity is the speed after the move. But if so, it really surprises me to initialize motion_velocity to self.vmax instead of 0.

python/lsst/ts/Dome/mock_llc_statuses/amcs_status.py
```diff
@@ -55,23 +54,34 @@ async def determine_status(self, current_tai):
...
+            if self.motion_velocity >= 0:
                 if self.position_actual >= self.position_cmd:
                     self.position_actual = self.position_cmd
-                    self.position_orig = self.position_actual
                     if self.status == LlcStatus.MOVING.value:
-                        self.status = LlcStatus.STOPPED.value
+                        if self.crawl_velocity >= 0:
+                            self.position_cmd = math.radians(720)
+                        else:
+                            self.position_cmd = math.radians(-360)
```
r-owen, line 65: Some explanation of these values would be appreciated. I think you mean these to be "always larger or smaller than the current position". If math.inf would work it might add clarity and eliminate a possible error if you leave the dome crawling for a long time (though you'd eventually run into inaccuracy due to rounding large angles).

python/lsst/ts/Dome/mock_llc_statuses/amcs_status.py
```diff
@@ -55,23 +54,34 @@ async def determine_status(self, current_tai):
...
+                        self.motion_velocity = self.crawl_velocity
+                        self.status = LlcStatus.CRAWLING.value
                     else:
+                        self.position_cmd = self.position_actual
+                        self.motion_velocity = 0
                         self.status = LlcStatus.PARKED.value
             else:
```
r-owen, line 72: Prior code, but I find the logic for changing to PARKED hard to follow and hard to trust. It seems to be "if not STOPPED and not MOVING and arrived then PARKED". That may be fine, but what if some new states are added, such as HOMING and HOMED?

I think it would be more robust to only transition to PARKED if the prior state was PARKING, and if you run out of options then raise an exception.

python/lsst/ts/Dome/mock_llc_statuses/amcs_status.py
```diff
@@ -55,23 +54,34 @@ async def determine_status(self, current_tai):
...
                     self.position_actual = self.position_cmd
-                    self.position_orig = self.position_actual
                     if self.status == LlcStatus.MOVING.value:
-                        self.status = LlcStatus.STOPPED.value
+                        if self.crawl_velocity >= 0:
+                            self.position_cmd = math.radians(720)
+                        else:
+                            self.position_cmd = math.radians(-360)
+                        self.motion_velocity = self.crawl_velocity
+                        self.status = LlcStatus.CRAWLING.value
```
r-owen, line 67: Prior code, but please consider setting self.status to the actual enum, instead of its value (e.g. `self.status = LlcStatus.CRAWLING`.

That simplifies the code, allows you to benefit from knowing status is an enum, not an arbitrary int, and if you ever print the value you get the name, instead of an integer that is hard to understand.

tests/test_mock_controller.py
```diff
@@ -303,7 +305,8 @@ async def test_stopEl(self):
 
     async def test_stop(self):
         target_azimuth = math.radians(10)
-        await self.write(f"moveAz:\n azimuth: {target_azimuth}\n")
+        target_rate = math.radians(0.1)
+        await self.write(f"moveAz:\n azimuth: {target_azimuth}\n azRate: {target_rate}")
```
r-owen, line 309: A central method to send commands would allow you to change the code in one place if the format changes again, and would make this code easier to follow -- i.e. just changing the existing `write` method to take a command and `**parameters`.

tests/test_mock_controller.py
```diff
@@ -126,15 +127,15 @@ async def test_moveAz(self):
         self.data = await self.read()
         amcs_status = self.data[Dome.LlcName.AMCS.value]
         self.assertEqual(
-            amcs_status["status"], Dome.LlcStatus.STOPPED.value,
+            amcs_status["status"], Dome.LlcStatus.CRAWLING.value,
         )
-        self.assertEqual(
+        self.assertGreaterEqual(
```
r-owen, line 132: Please also test moveAz and moveEl the crawl velocity 0 and negative.

Also I suggest testing elevation near the limits, to make sure it stops when it runs into a limit
and azimuth near the 0/360 boundary to make sure it crosses that boundary without a glitch.

The way I usually do this sort of thing is to write a `check_foo` method that has most of the logic, then call it from one or more `test_foo` methods that provide the conditions you are testing.

python/data/Dome_Commands.xml
```diff
@@ -12,6 +12,13 @@
         <Units>deg</Units>
         <Count>1</Count>
     </item>
+    <item>
+        <EFDB_Name>azRate</EFDB_Name>
+        <Description>Specifies the signed velocity of crawl motion in degrees per second at which the motion should continue once the commanded azimuth has been reached. Positive means in the direction of increasing azimuth, negative in the direction of decreasing azimuth.</Description>
```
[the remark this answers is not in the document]
wvreeven: I originally put them there as a reference for the external software manufacturer but we are sharing all info in another way now so I have removed the file.

python/lsst/ts/Dome/mock_controller.py
wvreeven: Done.

python/lsst/ts/Dome/mock_llc_statuses/amcs_status.py
wvreeven: Done

python/lsst/ts/Dome/mock_llc_statuses/amcs_status.py
wvreeven: I had math.inf (and -math.inf) before and decided to replace it. I don't remember why. I have put math.inf back with an inline comment.

python/lsst/ts/Dome/mock_llc_statuses/amcs_status.py
wvreeven: This evolved from the situation where there was no PARKING and PARKED yet and I agree that this could have been done in a better way. I followed your suggestion.

python/lsst/ts/Dome/mock_llc_statuses/amcs_status.py
wvreeven: The actual value is a string, not an integer. Having said that, I agree that this makes the code easier to read so I made the change in all status classes apart from the one for the louvers since numpy arrays don't seem to appreciate string-based enums.

tests/test_mock_controller.py
wvreeven: I was working on that but then this ticket gained priority so I parked the other one. See

https://jira.lsstcorp.org/browse/DM-25130

python/lsst/ts/Dome/mock_llc_statuses/amcs_status.py
wvreeven: On top of that, I really need to have the value of the Enum in the status string, and not the Enum itself. So even if the numpy array can hold Enums, I will need to add an additional step to get the values instead.

tests/test_mock_controller.py
wvreeven: I introduced test cases that test the 0 and negative velocities as well as moving to higher and lower azimuth values. I also opened a ticket to introduce proper handling of the 0/360 azimuth and 0/90 elevation boundaries:

https://jira.lsstcorp.org/browse/DM-25189

wvreeven:
Code changes pushed. Can you have another look please?

python/lsst/ts/Dome/mock_llc_statuses/amcs_status.py
```diff
@@ -53,28 +55,48 @@ async def determine_status(self, current_tai):
...
+                    # position should have been set to +/- math.inf
+                    raise ValueError(
+                        f"Went beyond the limit of {self.position_cmd} while in status {self.status.value}. "
+                        f"This should not happen."
+                    )
+                else:
+                    raise ValueError(f"Unknown state {self.status.value}")
         self.llc_status = {
-            "status": self.status,
+            "status": self.status.value,
```
r-owen, line 99: FYI: if you use IntEnum (I'm being too lazy to check, but consider it if you aren't) then I think you can skip the `.value`: it should "just work".

tests/test_mock_controller.py
```diff
@@ -79,62 +79,179 @@ async def test_missing_command_parameter(self):
         self.data = await self.read()
         self.assertEqual(self.data["ERROR"]["CODE"], 3)
 
-    async def test_moveAz(self):
-        target_azimuth = math.radians(10)
-        await self.write(f"moveAz:\n azimuth: {target_azimuth}\n")
+    async def prepare_amcs(
+        self, initial_position, target_azimuth, target_rate,
+    ):
+        """Utility method for preparing the in itial state of AMCS for easier testing.
```
r-owen, line 85: "in itial" -> "initial"

python/lsst/ts/Dome/mock_llc_statuses/amcs_status.py
r-owen: I am confused about this. Why use string values for enums? Ints should work just fine. When you print the enum  you get the name.

python/lsst/ts/Dome/mock_controller.py
```diff
@@ -247,13 +247,16 @@ async def move_az(self, **kwargs):
         ----------
         kwargs: `dict`
             A dictionary with arguments to the function call. It should contain the key "azimuth" with a
-            float value.
+            float value between 0 and 2pi and the key "azRate" with a float value where positive means
```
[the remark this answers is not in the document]
wvreeven: I added a comment to clarify this in

https://jira.lsstcorp.org/browse/DM-25189

python/lsst/ts/Dome/mock_controller.py
```diff
@@ -262,12 +265,12 @@ async def move_el(self, **kwargs):
         ----------
         kwargs: `dict`
             A dictionary with arguments to the function call. It should contain the key "elevation" with a
-            float value.
+            float value between 0 and pi/2.
```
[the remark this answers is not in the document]
wvreeven: I added a comment to clarify this in

https://jira.lsstcorp.org/browse/DM-25189

python/lsst/ts/Dome/mock_llc_statuses/amcs_status.py
wvreeven: The statuses sent by the LLCs contain strings like "Moving" and "Parked" and not "MOVING", "PARKED" or 3. So IntEnum will not work here.

tests/test_mock_controller.py
wvreeven: Thanks!

python/lsst/ts/Dome/mock_llc_statuses/amcs_status.py
r-owen: That is a fair point, though in fact enum values can be TitleCase (and if you only want the strings and not associated integers then IntEnum is irrelevant).

PR #26, opened 2020-06-16
Tickets/dm 25419

wvreeven:
Now all code changes are in place.

### DIFF
--- a/python/lsst/ts/Dome/dome_csc.py
+++ b/python/lsst/ts/Dome/dome_csc.py
@@ -20,7 +20,7 @@ _KEYS_TO_REMOVE = {"status"}
 _KEYS_IN_RADIANS = {"positionError", "positionActual", "positionCommanded"}
 
 _STATUS_TASK_PERIOD = 0.2
-_AMCS_PERIOD = 2.0
+_AMCS_PERIOD = 0.2
 _APsCS_PERIOD = 2.0
 _LCS_PERIOD = 2.0
 _LWSCS_PERIOD = 2.0
--- a/python/lsst/ts/Dome/dome_csc.py
+++ b/python/lsst/ts/Dome/dome_csc.py
@@ -5,8 +5,6 @@ import logging
 import math
 import pathlib
 
-import numpy as np
-
 from .llc_configuration_limits import AmcsLimits, LwscsLimits
 from .llc_name import LlcName
 from lsst.ts import salobj
@@ -19,13 +17,12 @@ _TIMEOUT = 20  # timeout in s to be used by this module
 _KEYS_TO_REMOVE = {"status"}
 _KEYS_IN_RADIANS = {"positionError", "positionActual", "positionCommanded"}
 
-_STATUS_TASK_PERIOD = 0.2
-_AMCS_PERIOD = 0.2
-_APsCS_PERIOD = 2.0
-_LCS_PERIOD = 2.0
-_LWSCS_PERIOD = 2.0
-_MonCS_PERIOD = 2.0
-_ThCS_PERIOD = 2.0
+_AMCS_STATUS_PERIOD = 0.2
+_APsCS_STATUS_PERIOD = 2.0
+_LCS_STATUS_PERIOD = 2.0
+_LWSCS_STATUS_PERIOD = 2.0
+_MONCS_STATUS_PERIOD = 2.0
+_THCS_STATUS_PERIOD = 2.0
 
 
 class DomeCsc(salobj.ConfigurableCsc):
@@ -58,10 +55,9 @@ class DomeCsc(salobj.ConfigurableCsc):
         self.writer = None
         self.config = None
 
-        self.log = logging.getLogger("DomeCsc")
-
         self.mock_ctrl = None  # mock controller, or None if not constructed
         self.mock_port = mock_port  # mock port, or None if not used
+
         super().__init__(
             name="Dome",
             index=0,
@@ -71,9 +67,15 @@ class DomeCsc(salobj.ConfigurableCsc):
             simulation_mode=simulation_mode,
         )
 
+        stream_handler = logging.StreamHandler()
+        self.log.addHandler(stream_handler)
+
         # Keep the lower level statuses in memory for unit tests.
         self.lower_level_status = {}
-        self.llc_status_task = None
+        self.status_tasks = []
+
+        # Keep a lock so only one remote command can be executed at a time.
+        self.communication_lock = asyncio.Lock()
 
         self.amcs_limits = AmcsLimits()
         self.lwscs_limits = LwscsLimits()
@@ -104,12 +106,31 @@ class DomeCsc(salobj.ConfigurableCsc):
             connect_coro, timeout=self.config.connection_timeout
         )
 
-        # Start polling for the status of the AMCS component periodically at a frequency of 5 Hz.
-        self.llc_status_task = asyncio.create_task(
-            self.schedule_llc_tasks_periodically()
-        )
+        # Start polling for the status of the lower level components periodically.
+        await self.start_status_tasks()
 
         self.log.info("connected")
+
+    async def cancel_status_tasks(self):
+        """Cancel all status tasks."""
+        while self.status_tasks:
+            self.status_tasks.pop().cancel()
+
+    async def start_status_tasks(self):
+        """Start all status tasks
+        """
+        await self.cancel_status_tasks()
+        for method, interval in (
+            (self.statusAMCS, _AMCS_STATUS_PERIOD),
+            (self.statusApSCS, _APsCS_STATUS_PERIOD),
+            (self.statusLCS, _LCS_STATUS_PERIOD),
+            (self.statusLWSCS, _LWSCS_STATUS_PERIOD),
+            (self.statusMonCS, _MONCS_STATUS_PERIOD),
+            (self.statusThCS, _THCS_STATUS_PERIOD),
+        ):
+            self.status_tasks.append(
+                asyncio.create_task(self.one_status_loop(method, interval))
+            )
 
     async def disconnect(self):
         """Disconnect from the TCP/IP controller, if connected, and stop the mock controller, if running.
@@ -117,8 +138,7 @@ class DomeCsc(salobj.ConfigurableCsc):
         self.log.info("disconnect")
 
         # Stop polling for the status of the lower level components periodically.
-        if self.llc_status_task:
-            self.llc_status_task.cancel()
+        await self.cancel_status_tasks()
 
         writer = self.writer
         self.reader = None
@@ -188,21 +208,26 @@ class DomeCsc(salobj.ConfigurableCsc):
          """
         cmd = params["command"]
         st = encoding_tools.encode(**params)
-        self.log.debug(f"Sending command {st}")
-        self.writer.write(st.encode() + b"\r\n")
-        await self.writer.drain()
-        read_bytes = await asyncio.wait_for(self.reader.readuntil(b"\r\n"), timeout=1)
-        data = encoding_tools.decode(read_bytes.decode())
+        async with self.communication_lock:
+            self.log.debug(f"Sending command {st}")
+            self.writer.write(st.encode() + b"\r\n")
+            await self.writer.drain()
+            read_bytes = await asyncio.wait_for(
+                self.reader.readuntil(b"\r\n"), timeout=1
+            )
+            data = encoding_tools.decode(read_bytes.decode())
 
-        response = data["response"]
-        if response > ResponseCode.OK:
-            self.log.error(f"Received ERROR {data}.")
-            if response == ResponseCode.INCORRECT_PARAMETER:
-                raise ValueError(f"The command {cmd} contains an incorrect parameter.")
-            elif response == ResponseCode.UNSUPPORTED_COMMAND:
-                raise KeyError(f"The command {cmd} is unsupported.")
+            response = data["response"]
+            if response > ResponseCode.OK:
+                self.log.error(f"Received ERROR {data}.")
+                if response == ResponseCode.INCORRECT_PARAMETER:
+                    raise ValueError(
+                        f"The command {cmd} contains an incorrect parameter."
+                    )
+                elif response == ResponseCode.UNSUPPORTED_COMMAND:
+                    raise KeyError(f"The command {cmd} is unsupported.")
 
-        return data
+            return data
 
     async def do_moveAz(self, data):
         """Move AZ.
@@ -214,7 +239,7 @@ class DomeCsc(salobj.ConfigurableCsc):
         """
         self.assert_enabled()
         self.log.debug(
-            f"Moving Dome to azimuth {data.position} and gthen start crawling at azRate {data.velocity}"
+            f"Moving Dome to azimuth {data.position} and then start crawling at azRate {data.velocity}"
         )
         await self.write_then_read_reply(
             command="moveAz",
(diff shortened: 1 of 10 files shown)


python/lsst/ts/Dome/dome_csc.py
```diff
@@ -598,21 +600,37 @@ async def implement_simulation_mode(self, simulation_mode):
...
+                    reset_start_tai = True
+                if (current_tai - start_tai) > _MonCS_PERIOD:
+                    await self.statusMonCS()
+                    reset_start_tai = True
+                if (current_tai - start_tai) > _ThCS_PERIOD:
+                    await self.statusThCS()
+                    reset_start_tai = True
+
+                if reset_start_tai:
+                    start_tai = current_tai
```
r-owen, line 631: I strongly suspect this code is broken. The problem is that you reset start_tai if *any* status is output, which means that any status with a longer period will never run.

You could keep a separate start time for each method, but that's a lot of state to keep track of and a lot of repetition. Also this design means you can't set truly independent intervals for the various kinds of status.

Consider running a separate loop for each status. If you hold them in a collection then they are easy to cancel when shutting down or go out of ENABLED/DISABLED.
```
def __init__(self, ...):
    self.status_tasks = []

async def one_status_loop(self, method, interval):
    """Run one status method forever at the specified interval.
    ...
    """
    try:
        while True:
            method()
            await asyncio.sleep(interval)
    except Exception:
        self.log.exception(f"one_status_loop({method}) failed")

async def cancel_status_tasks(self):
    "Cancel all status tasks."""
    while self.status_tasks:
        self.status_tasks.pop().cancel()

async def start_status_tasks(self):
    """Start all status tasks
    """
    self.cancel_status_tasks()
    for method, interval in ((self.statusAMCS, AMCS_STATUS_INTERVAL)), ...):
        self.status_tasks.append(asyncio.create_task(self.one_status_loop(method, interval))
```
You could also make start_status_tasks wait for all the tasks to finish, in which case you could just cancel the task that runs it to shut down all status. But you have to be a bit careful in your use of `gather` so that if one status task fails it doesn't leave the main task dead and a bunch of subtasks running (makes for a lot of warnings at shutdown). But I think the above is a bit simpler, and it assures that all status tasks are canceled.

python/lsst/ts/Dome/mock_controller.py
```diff
@@ -246,6 +246,9 @@ async def request_and_send_status(self, llc, llc_name):
         llc_name: LlcName
             The name of the Lower Level Component.
         """
+        self.log.debug("Determining current TAI.")
+        await self.determine_current_tai()
```
r-owen, line 250: Where this method is defined (a few lines past this block of changess): please modify the doc string:
* Say that it returns the current TAI (right now it talks about a time difference, which is no longer true)
* The method exists so you can replace it with a mock for unit tests.

(This assumes you are still finding it useful to mock the method. If not, please consider getting rid of it.)

python/lsst/ts/Dome/mock_controller.py
```diff
@@ -261,16 +264,16 @@ async def move_az(self, **kwargs):
         Parameters
         ----------
         kwargs: `dict`
-            A dictionary with arguments to the function call. It should contain the key "azimuth" with a
-            float value between 0 and 2pi and the key "azRate" with a float value where positive means
+            A dictionary with arguments to the function call. It should contain the key "position" with a
+            float value between 0 and 2pi and the key "velocity" with a float value where positive means
```
r-owen, line 268: For this and all other commands please list the arguments directly instead of using `kwargs`. In other words:
```
async def move_az(self, position, velocity):
    """Move the dome.

    Parameters
    -----------
    position : `float`
        Desired azimuth, in radians, in range [0, 2 pi)
    ...
    """
```
It makes the API much clearer and easier to understand and catches mistakes in spelling the arguments immediately.

The only thing it makes harder is logging the info, and I personally think a clear API is much more important than simplicity of logging.

python/lsst/ts/Dome/mock_llc_statuses/amcs_status.py
```diff
@@ -37,14 +37,14 @@ def __init__(self):
...
+        self.position_commanded = 0
+        self.velocity_actual = 0
+        self.velocity_commanded = 0
         self.drive_torque_actual = np.zeros(_NUM_MOTORS, dtype=float)
-        self.drive_torque_error = np.zeros(_NUM_MOTORS, dtype=float)
-        self.drive_torque_cmd = np.zeros(_NUM_MOTORS, dtype=float)
+        self.drive_torque_commanded = np.zeros(_NUM_MOTORS, dtype=float)
         self.drive_current_actual = np.zeros(_NUM_MOTORS, dtype=float)
-        self.drive_temp_actual = np.full(_NUM_MOTORS, 20.0, dtype=float)
+        self.drive_temperature = np.full(_NUM_MOTORS, 20.0, dtype=float)
```
r-owen, line 47: Nice. I didn't know about `full`

python/lsst/ts/Dome/mock_controller.py
```diff
@@ -338,24 +341,20 @@ async def crawlEl(self, **kwargs):
 
         # No conversion from radians to degrees needed since both the commands and the mock az controller
         # use radians.
-        await self.lwscs.crawlEl(velocity=float(kwargs["elRate"]))
+        await self.lwscs.crawlEl(velocity=float(kwargs["velocity"]))
 
-    async def setLouver(self, **kwargs):
-        """Set the position of a louver.
+    async def setLouvers(self, **kwargs):
```
r-owen, line 346: Picky point but... you have `move_az` but `setLouvers` in this class and `moveAz` in another class. If there isn't some underlying logic that drives these choices please consider standardizing. It would make your already clear code even easier to follow.

python/lsst/ts/Dome/mock_llc_statuses/amcs_status.py
```diff
@@ -64,77 +64,84 @@ async def determine_status(self, current_tai):
...
-    async def moveAz(self, azimuth, velocity):
+    async def moveAz(self, position, velocity):
         """Move the dome at maximum velocity to the specified azimuth. Azimuth is measured from 0 at
             north via 90 at east and 180 at south to 270 west and 360 = 0. The value of azimuth is not
             checked for the range between 0 and 360.
 
         Parameters
         ----------
-        azimuth: `float`
+        position: `float`
```
r-owen, line 131: Please document the units.

tests/test_mock_controller.py
```diff
@@ -3,6 +3,7 @@
 from asynctest.mock import CoroutineMock
 import logging
 import math
+import numpy as np
```
r-owen, line 6: Picky point, but please insert a blank line between standard library and 3rd party imports

tests/test_mock_controller.py
```diff
@@ -82,17 +83,17 @@ async def test_missing_command_parameter(self):
         self.assertEqual(self.data["timeout"], -1)
 
     async def prepare_amcs(
-        self, initial_position, target_azimuth, target_rate,
+        self, initial_position, target_position, target_velocity,
     ):
         """Utility method for preparing the in initial state of AMCS for easier testing.
```
r-owen, line 88: "preparing the in initial state"? Is there an extra "in"? If not, might you reword for clarity?

tests/test_mock_controller.py
```diff
@@ -554,27 +573,45 @@ async def test_setLouver(self):
...
-            + [target_position]
-            + [0.0] * (_NUM_LOUVERS - louver_id - 1),
-        )
-        self.assertEqual(
-            lcs_status["positionCmd"],
-            [0.0] * louver_id
-            + [target_position]
-            + [0.0] * (_NUM_LOUVERS - louver_id - 1),
-        )
+        lcs_status = self.data[LlcName.LCS.value][0]
```
r-owen, line 579: A code comment describing the status would be a big help here, as well as a doc string.
I am not sure what `louver_ids` is (well, reading on I see it's a list of louver ID numbers, which makes sense).

What does lcs_status["status"] look like, and thus what the value of status might be inside the loop?
Why does a non-zero target_position imply OPEN? You don't set all positions to 100, so they aren't fully open. I would naively have guessed that they would be neither OPEN nor CLOSED unless they had value 100 or 0.

python/lsst/ts/Dome/dome_csc.py
```diff
@@ -465,61 +464,64 @@ async def statusAMCS(self):
...
     async def request_and_send_llc_status(self, llc_name, telemetry_function):
         command = f"status{llc_name.value}"
         status = await self.write_then_read_reply(command=command, parameters={})
         # Store the status for unit tests.
-        self.lower_level_status[llc_name.value] = status[llc_name.value]
-        self.log.info(status)
-        self.convert_telemetry_radians_to_degrees_and_send(
-            status[llc_name.value], telemetry_function, llc_name
-        )
+        self.lower_level_status[llc_name.value] = status[llc_name.value][0]
```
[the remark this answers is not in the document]
wvreeven: Since there is no conversion to radians involved with the light and wind screen, I have removed the convert_telemetry_list_radians_to_degrees_and_send method and simplified the code quite a bit.

python/lsst/ts/Dome/dome_csc.py
wvreeven: Thanks very much for the suggestion. I adopted it and it works like a charm.

python/lsst/ts/Dome/mock_controller.py
wvreeven: Thanks for being picky!

tests/test_mock_controller.py
wvreeven: Thanks, I keep forgetting about that.

tests/test_mock_controller.py
wvreeven: I added some doc string to the lcs_status class and an inline comment just above the line you mentioned.

python/lsst/ts/Dome/mock_llc_statuses/amcs_status.py
wvreeven: Done

python/lsst/ts/Dome/mock_controller.py
wvreeven: Done

tests/test_mock_controller.py
wvreeven: Done

tests/test_mock_controller.py
wvreeven: Extra "in" removed.

python/lsst/ts/Dome/dome_csc.py
```diff
@@ -50,10 +55,9 @@ def __init__(
         self.writer = None
         self.config = None
 
-        self.log = logging.getLogger("DomeCsc")
```
r-owen, line 53: Thank you.

python/lsst/ts/Dome/dome_csc.py
```diff
@@ -63,9 +67,15 @@ def __init__(
             simulation_mode=simulation_mode,
         )
 
+        stream_handler = logging.StreamHandler()
+        self.log.addHandler(stream_handler)
```
r-owen, line 71: It is very unusual for a CSC to stream log messages. Is this temporary? If it's permanent then please consider RotatingFileHandler or SysLogHandler.

python/lsst/ts/Dome/dome_csc.py
```diff
@@ -188,21 +208,26 @@ async def write_then_read_reply(self, **params):
...
-            elif response == ResponseCode.UNSUPPORTED_COMMAND:
-                raise KeyError(f"The command {cmd} is unsupported.")
-
-        return data
+        async with self.communication_lock:
+            self.log.debug(f"Sending command {st}")
+            self.writer.write(st.encode() + b"\r\n")
+            await self.writer.drain()
+            read_bytes = await asyncio.wait_for(
+                self.reader.readuntil(b"\r\n"), timeout=1
```
r-owen, line 216: Please consider a named constants for the timeout, so it's visible near the top of the file (with the other constants).

python/lsst/ts/Dome/dome_csc.py
```diff
@@ -213,14 +238,14 @@ async def do_moveAz(self, data):
...
+        self.log.debug(
+            f"Moving Dome to azimuth {data.position} and then start crawling at azRate {data.velocity}"
         )
         await self.write_then_read_reply(
             command="moveAz",
             parameters={
-                "azimuth": math.radians(data.azimuth),
-                "azRate": math.radians(data.azRate),
+                "position": math.radians(data.position),
+                "velocity": math.radians(data.velocity),
```
r-owen, line 248: I strongly recommend having `write_then_read_reply` take command and ``**kwargs`` as parameters. That  clarifies the API, makes it easier to call correctly and harder to call incorrectly. For example:
```
    await self.write_then_read_reply("moveEl", position=math.radians(data.position))
    # or, if you prefer (both work fine):
    await self.write_then_read_reply(command="moveEl", position=math.radians(data.position))
```
Where the method is implemented as something like:
```
async def write_then_read_reply(self, command, **params):
    command_dict = dict(command=command, parameters=params)
    command_encoded = encoding_tools.encode(**command_dict)
```

python/lsst/ts/Dome/dome_csc.py
```diff
@@ -399,35 +423,27 @@ async def config_llcs(self, data):
         Parameters
         ----------
         data : `dict`
-            A dictionary with arguments to the function call. It should contain keys for all lower level
-            components to be configured with values that are dicts with keys for all the parameters that
-            need to be configured. The structure is::
+            A dictionary with arguments to the function call. It should contain a key "system" indicating
```
r-owen, line 426: I strongly recommend changing this to an API that allows specifying the "system" and "settings" as arguments:
```
async def config_llcs(self, system, settings):
```
That allows you to document the actual arguments, Python checks that they are spelled correctly, and the user does not have to needlessly make an extra struct to call this.

python/lsst/ts/Dome/dome_csc.py
```diff
@@ -465,86 +481,73 @@ async def statusAMCS(self):
...
-    def convert_telemetry_radians_to_degrees_and_send(
-        self, telemetry_in_radians, telemetry_function, llc_name
-    ):
         telemetry_in_degrees = {}
+        telemetry_in_radians = status[llc_name.value][0]
         for key in telemetry_in_radians.keys():
-            if key in _KEYS_IN_RADIANS and (
-                llc_name == LlcName.AMCS or llc_name == LlcName.APSCS
-            ):
+            # LWSCS is handled by convert_telemetry_list_radians_to_degrees_and_send
```
r-owen, line 541: I think you got rid of `convert_telemetry_list_radians_to_degrees_and_send`? If so, please update or delete this comment. I confess I'm not sure what it is telling me.

python/lsst/ts/Dome/mock_controller.py
```diff
@@ -246,224 +238,225 @@ async def request_and_send_status(self, llc, llc_name):
...
+        self.log.debug("Determining current TAI.")
+        await self.determine_current_tai()
+        self.log.debug(f"Requesting status for LLC {llc_name}")
         await llc.determine_status(self.current_tai)
         state = {llc_name.value: llc.llc_status}
         await self.write(response=ResponseCode.OK, **state)
 
     async def determine_current_tai(self):
-        """Determine the time difference since the previous call.
+        """Determine the cureent TAI time.
```
r-owen, line 249: cureent -> current

python/lsst/ts/Dome/mock_controller.py
```diff
@@ -246,224 +238,225 @@ async def request_and_send_status(self, llc, llc_name):
...
-                    "amax"
-                    "vmax"
-                "LWSCS":
-                    "jmax"
-                    "amax"
-                    "vmax"
+        system: `str`
+            The name of the system to configure.
+        settings: `list` of `dict`
+            A dictionary with arguments to the function call. It should contain a key called "system" with
```
r-owen, line 387: I believe this comment needs to be updated. I think the first two sentences can be deleted, so the real description starts around "an array containing a dict".

Also please consider "an array containing a single dict" to emphasize that we really only expect one dict. (I realize this is the way EIE does it, and it's fine, but it is likely to surprise anyone who thinks you designed this API).

python/lsst/ts/Dome/dome_csc.py
wvreeven: It is not temporary and this actually was recommended by Tiago. I do agree on using a RotatingFileHandler so I've changed to that.

python/lsst/ts/Dome/dome_csc.py
wvreeven: There already was one so I am using that one now.

python/lsst/ts/Dome/dome_csc.py
wvreeven: Thanks for catching that! I removed the comment.

python/lsst/ts/Dome/mock_controller.py
wvreeven: Corrected.

python/lsst/ts/Dome/mock_controller.py
wvreeven: Corrected, thanks!

python/lsst/ts/Dome/dome_csc.py
wvreeven: There are two ways to do this: make settings a dict or make settings an array containing a single dict. I chose the first to keep things simple on the CSC side and up.

wvreeven:
New changes pushed.

python/lsst/ts/Dome/dome_csc.py
```diff
@@ -387,47 +405,34 @@ async def do_setTemperature(self, data):
...
+            A dict containing key,value for all the parameters that need to be configured. The structure is::
 
-        It is assumed that configuration_parameters is presented as a dictionary of dictionaries with one
-        dictionary per lower level component. This means that we only need to check for unknown and too
-        large parameters and then send all to the lower level components. An example would be::
-
-            {"AMCS": {"amax": 5, "jmax": 4}, "LWSCS": {"vmax": 5, "jmax": 4}}
+                "jmax"
+                "amax"
+                "vmax"
```
r-owen, line 426: Minor nit, but if you *know* that it's always jmax, amax, vmax, then it would be clearer to have those be individual arguments to this method. On the other hand if this is just an example, then your API is great, and it'd be helpful to note that this is merely an example.

wvreeven: For now it's only those parameters but there will be (possibly many) more in the future so I'll keep it like this. Also, the other systems will very likely get config params as well.

PR #30, opened 2020-07-09
Replaced all list-tables with csv-tables.

wvreeven:


wvreeven:
Thanks. Yes I noticed that too. No idea why though. Probably some change in a TSSW project that ts_Dome depends on.

PR #31, opened 2020-09-16
Reformat code with black.

tribeiro:
Fix f-strings with no string substitution.
Pin black version in meta.yaml to 19.10b0

### DIFF
--- a/conda/meta.yaml
+++ b/conda/meta.yaml
@@ -35,7 +35,7 @@ test:
     - ts-dds
     - ts-idl
     - ts-salobj
-    - black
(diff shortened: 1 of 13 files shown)


wvreeven (changes requested):
In principles the changes look good. However, the project already was formatted with black but with a line length of 110 characters (except the doc and comment lines) instead of the black default of 88. Can you take care of that please?

wvreeven (approved):
After a discussion with the TSSW team we decided to stick with the default black line length of 88 characters.

PR #32, opened 2020-09-23
Made changes regarding the statuses of the lower level components and…

wvreeven:
… the configuration protocol.

### DIFF
--- a/python/lsst/ts/Dome/dome_csc.py
+++ b/python/lsst/ts/Dome/dome_csc.py
@@ -63,7 +63,18 @@ class DomeCsc(salobj.ConfigurableCsc):
         * 1: simulation: use a mock low level HVAC controller.
     mock_port : `int`
         The port that the mock controller will listen on
+
+    Notes
+    -----
+    **Simulation Modes**
+
+    Supported simulation modes:
+
+    * 0: regular operation
+    * 1: simulation mode: start a mock TCP/IP Dome controller and talk to it
     """
+
+    valid_simulation_modes = (0, 1)
 
     def __init__(
         self,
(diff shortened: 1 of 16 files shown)


python/lsst/ts/Dome/dome_csc.py
```diff
@@ -68,9 +72,10 @@ def __init__(
         simulation_mode=0,
         mock_port=None,
     ):
-        schema_path = (
-            pathlib.Path(__file__).resolve().parents[4].joinpath("schema", "Dome.yaml")
-        )
+        if simulation_mode not in (0, 1):
```
couger01, line 75: This is being deprecated in salobj 6, the new way is described [here](https://ts-salobj.lsst.io/v/develop/salobj_cscs.html#simulation-mode).

couger01 (changes requested):
Only one change to consider. Otherwise LGTM.

python/lsst/ts/Dome/dome_csc.py
wvreeven: Thanks for noticing that. I am afraid there is a catch 22 here. The current development docker image still uses salobj 5 which is also specified in the version history. If I change this now then the unit tests will fail, both locally and in Jenkins. However, making the change now would prepare it for salobj6 but there is no (easy) way for me to test that the change will work.
Any idea when salobj 6 will be released? Would you accept it if I promised to make the change you suggested as soon as that release is done? I will create a new Jira ticket for it.

couger01: The develop-env:develop is running the develop branch of salobj so that should be fine for jenkins although I see your Jenkinsfile is running salobj:develop.

wvreeven: Changes made and tested against develop-env:develop. I also made sure that Jenkins uses that docker image.

couger01 (approved):
LGTM

PR #37, opened 2020-11-03
Renamed Dome to MTDome.

wvreeven:


r-owen (changes requested):
Overall this looks great. Three requests:
* Please update labels=Dome and the associated URL in index.rst. We usually use the package name: `ts_MTDome`. That avoids confusion with more generic terms.
* Please change "Dome CSC" to "MTDome CSC" in base_llc_motion.py .
* Please rename the labels in the existing Jira tickets for your software to use the new name. Normally I'd suggest Jira's support for mass changes, but in this case it appears there are a number of tickets with the label Dome that aren't necessarily about your package so it may have to be more manual.

You have bare Dome references in a few other mock controller files that _could_ also be changed to MTDome, but the docs are already clear.

wvreeven:
I addressed all three points and even changed all other Dome references to MTDome.

r-owen (approved):
Thank you for the changes. This is good to go.

PR #38, opened 2020-11-20
DM-27664: Switched to pre-commit, ts-conda-build and JenkinsShared and small co…

wvreeven:
…de changes due to changes in ts_xml for MTDome.


conda/meta.yaml
```diff
@@ -56,5 +44,5 @@ requirements:
     - python
     - setuptools
     - setuptools_scm
-    - ts-salobj
-    - ts-simactuators
+    - ts-idl {{ idl_version }}
+    - ts-salobj {{ salobj_version }}
```
r-owen, line 48: Can you tell me where these values are set? I'm just curious.

Jenkinsfile.conda
```diff
@@ -1,72 +1,2 @@
...
-    post {
-        cleanup {
-            sh """
-            docker stop ${container_name}
-            """
-        }
-    }
-}
+@Library('JenkinsShared')_
+CondaPipeline(["ts_config_mttcs"], "ts-mtdome", "lsst.ts.mtdome")
```
r-owen, line 2: Nice! I hope we get something like this for Jenkinsfile as well.

doc/version_history.rst
```diff
@@ -6,6 +6,21 @@
 Version History
 ###############
 
+v0.6.0
+======
+
+| Switched to pre-commit
+| Switched to ts-conda-build
+| Switched to JenkinsShared for the conda build
+| Small code updates due to modifications in ts_xml for MTDome
```
r-owen, line 15: What does the `|` do? Please consider * instead of | for consistency.

I wonder if we can start dropping the Requires section from our version histories, if the dependencies are clear enough in another file (but which one?). Not sure we're there yet, but looking forward to it.

python/lsst/ts/MTDome/mock_controller.py
```diff
@@ -483,7 +483,7 @@ async def config(self, system, settings):
             # DM-26653: The structure of the settings variable has changed so
             # we need slightly more complex code here. For info about the new
             # structure, see the project docs at
-            # https://ts-dome.lsst.io/protocols.html#configuration-protocol
+            # https://ts-mtdome.lsst.io/protocols.html#configuration-protocol
```
r-owen, line 486: The ticket DM-26653 referenced above has already been marked Done. But a ticket number sounds like a "to do".

I'm not convinced you need any comment here (configuration is documented in the schema itself and in the standard place in the CSC docs). But if you really need one I'd stick to "For information about the configuration see https://..." or similar.

For future reference: when the config schema changes in a non-backwards-compatible way you should update the version number (in the title and correspondingly in the mt_config_xxx package), but I am not suggesting you do that now and it's probably not crucial until the package is released.

python/lsst/ts/MTDome/mock_llc/amcs.py
```diff
@@ -33,6 +33,8 @@
 from ..on_off import OnOff
 
 _NUM_MOTORS = 5
+_NUM_ENCODERS = 5
+_NUM_RESOLVERS = 3
```
r-owen, line 37: Good addition.

r-owen: I wonder if you should make all these public, to aid unit testing? But perhaps there are easy ways to get these values.

python/lsst/ts/MTDome/mock_llc/thcs.py
```diff
@@ -40,7 +40,7 @@ def __init__(self):
         self.log = logging.getLogger("MockThcsStatus")
         # variables holding the status of the mock Louvres
         self.status = MotionState.CLOSED
-        self.temperature = np.zeros(_NUM_SENSORS, dtype=float)
+        self.temperature = np.zeros(NUM_THERMO_SENSORS, dtype=float)
```
r-owen, line 43: The comment "# variables holding ..." (a few lines above) looks out of place here. Please consider deleting it.

tests/test_mock_controller.py
```diff
@@ -717,7 +717,7 @@ async def prepare_louvers(self, louver_ids, target_positions):
         target positions are lined up.
 
         """
-        position = np.full(_NUM_LOUVERS, -1.0, dtype=float)
+        position = np.full(NUM_LOUVERS, -1.0, dtype=float)
```
r-owen, line 720: What is special about -1? Is it "do not change this value"?

r-owen: I have checked out the code and read it and I don't understand what this is doing. You go to a lot of trouble to move louver 5 to 100 and the rest to -1, but I don't see anywhere that you check that what you requested happens.

If this was a way of cycling through the louvers (just move one or a few at a time) I suggest a "check_move_louver" method that takes the identify of a lover to move and moves it (or even a list of louvers to move) and verifies that it/they move as expected. Then you can have a test_move_louver command that calls check_move_louver once per louver (or moves several collections, or even moves all possible combinations, depending on how many there are and how long the check method takes).

r-owen (approved):
The advertised change looks great.

I am a bit concerned that the code that tests louver motion is not being sufficiently careful to actually check that the specified louvers moved where they were told to go, but I may well be missing something, Please have a look and if I'm right please consider beefing up that test before merging.

conda/meta.yaml
wvreeven: I suppose somewhere in the Docker environment. Rob asked me to add that so I did :)

Jenkinsfile.conda
wvreeven: Me too.

doc/version_history.rst
wvreeven: Not sure where I got that from. I have changed them all to *.

python/lsst/ts/MTDome/mock_controller.py
wvreeven: Thanks. I added the ticket reference so it was clear where the modification was introduced. But I agree that that is open to interpretation so I removed the entire comment. Thanks for informing me about updating the version of the schema.

python/lsst/ts/MTDome/mock_llc/amcs.py
wvreeven: I wondered about it myself as well and decided to do that once the unit tests actually use them, which is not the case at this moment. It is work in progress.

python/lsst/ts/MTDome/mock_llc/thcs.py
wvreeven: Nice catch. It was yet another case of bad copying and pasting. I removed the comments as you suggested.

tests/test_mock_controller.py
wvreeven: As a matter of fact, you suggested several months that -1 indeed means "do not move". I agree that the code is complex. My intention was to move only a few and to check that they indeed move. I agree that the code can be simplified. I'll consider improving the code once EIE finally has gotten around implementing the louvers control on their side because I am sure that my side will need changes as well.

PR #41, opened 2020-11-24
DM-27703: Fixed Jenkinsfile to the documentation gets published.

wvreeven:


Jenkinsfile
```diff
@@ -83,37 +84,47 @@ pipeline {
...
+                reportDir: 'jenkinsReport/',
+                reportFiles: 'index.html',
+                reportName: "Coverage Report"
+              ])
+
             sh "docker exec -u saluser \${container_name} sh -c \"" +
                 "source ~/.setup.sh && " +
                 "cd /home/saluser/repo/ && " +
-                "setup ts_MTDome -t saluser && " +
+                "setup ts_salobj -t saluser && " +
```
r-owen, line 108: I'm a bit confused here. Don't you need to be in the ts_MTDome directory and have it setup before you build its docs? Yet you are setting up ts_salobj, not ts_MTMount.

r-owen: What happens if `package-docs build` fails? I hope it doesn't go on and try to upload the docs. Naively I would have expected this to be a script, and call unstable if the result is not 0. I see ts_salobj does things the same way. I wonder if that means it needs updating as well?

r-owen (approved):
I have two questions I hope you will look at before merging. Assuming it works, it's good to go.

Jenkinsfile
wvreeven: Thanks for catching that. I basically copied that part from the ts_salobj Jenkinsfile and forgot to modify that line, which I will do now.

wvreeven: Also, this id done from the `repo` directory and not `repos`.

PR #43, opened 2021-01-07
DM-27821: Added validation of outgoing and incoming JSON data based o…

wvreeven:
…n JSON schemas.


python/lsst/ts/MTDome/encoding_tools.py
```diff
@@ -57,4 +756,60 @@ def decode(st):
...
+    Raises
+    ------
+    ValidationError:
+        In case the validation fails.
+    ValueError:
+        In case the retrieved data doesn't contain a known key.
+
+    """
+    try:
+        if "command" in data.keys():
```
pkubanek, line 789: This code can be significantly simplified. There are a few things I would like to change:

1. put your schema into files. Load them in module (not in any function), such as:
```python
command_schema = json.loadstr(open('command_schema.jschema').read())
```

2. this long if/elif can be settled with a hash you will intialise. Keys would be commands, values schemas. So:
```python
encodings = { 'command' : command_schema, .. 'ThCs' : tchs_status_schema }
```
or even better - if you agree to keep schemas in files, just put filename as value and use something like:
```python
encoding = {k: json.load(open(v).read()) for k, v in encoding_filenames.items()}
```

3. trivially, after you have dictionary where keys are commands and values are schemas, do something like:
```python
for k,v in encoding.items():
     if k in data.keys():
          jsonschema.validate(data, v)

if validation_raises_exception:
    raise RuntimeError(f"Validation failed because no known key found in data {data!r}")

File: python/lsst/ts/MTDome/encoding_tools.py [hunk 2]
```diff
@@ -57,4 +756,60 @@ def decode(st):
...
+        elif "LWSCS" in data.keys():
+            jsonschema.validate(data, lwscs_status_schema)
+        elif "MonCS" in data.keys():
+            jsonschema.validate(data, moncs_status_schema)
+        elif "ThCS" in data.keys():
+            jsonschema.validate(data, thcs_status_schema)
+        else:
+            message = f"Validation failed because no known key found in data {data!r}"
+            if validation_raises_exception:
+                raise KeyError(message)
```
pkubanek, line 808: KeyError is a bit misleading. I understand the logic, but KeyError is kind-of reserved for dict. Better use RuntimeError(message), so if true KeyError happens, you will not get it..

tests/test_mock_controller.py
```diff
@@ -97,18 +99,30 @@ async def tearDown(self):
             self.writer.close()
 
     async def test_command_does_not_exist(self):
+        # Temporarily disable validation exceptions for the unit test.
+        # Validation of the commands should be done by the client and the
+        # simulator has such validation built in.
+        MTDome.encoding_tools.validation_raises_exception = False
```
pkubanek, line 105: Do you really need this? It might be better to keep it True

doc/protocols.rst
```diff
@@ -146,3 +155,731 @@ If during the timeout another command except the status command is received then
...
+
+This is the command schema.
+It looks rather complex since it covers all available commands (including the "config" command), however otherwise 24 separate schemas would need to be created.
+For a full list of the commands and their parameters, see `Lower Level Commands`_.
+
+.. _Lower Level Commands: ./commands.html
+
+.. code-block:: none
+
+    {
```
pkubanek, line 178: See comment on validation_tools. When you keep schemas in files, you can put links there to schema files instead of copy&paste.

pkubanek (changes requested):
Please see comments.

python/lsst/ts/MTDome/encoding_tools.py
wvreeven: Thanks for the great suggestion. I put the schemas in files and very much simplified determining which schema to use.

python/lsst/ts/MTDome/encoding_tools.py
wvreeven: Thanks, I modified that.

tests/test_mock_controller.py
wvreeven: In the three particular test cases where I added that line, it indeed is needed since the simulator needs to do the validation and not the encoding tools, as I hoped the comment would explain.

doc/protocols.rst
wvreeven: As discussed on Slack yesterday, I will modify this as well.

wvreeven:
Changes pushed. Can you please have another look?

python/lsst/ts/MTDome/encoding_tools.py
```diff
@@ -19,7 +19,57 @@
...
+)
+moncs_status_schema = json.loads(
+    open(os.path.join(script_dir, "schemas/moncs_status.jschema")).read()
+)
+thcs_status_schema = json.loads(
+    open(os.path.join(script_dir, "schemas/thcs_status.jschema")).read()
+)
+
+# dict to help look up the schema to use for validation
+schemas = {
```
pkubanek, line 56: This can be a bit improved - the advantage would be that you can get easily error handling if needed. Instead of _schema = json.loads(...), do:

```python
def open_schema(name):
    return json.loads(open(os.path.join(script_dir, "schemas", filename)).read())
```

and then:

```python
   "command": open_schema("command.jschema")
```

etc.

pkubanek (approved):
Much appreciated changes. Please see comment on how to use a function to further reduce number of command lines (and allow for better handling of errors).

PR #47, opened 2021-02-02
DM-28436: Removed several obsolete test cases and updated the version…

wvreeven:
… history to include all missing tags.


andrewbheyer (approved):
Wouter improved his tests by cleaning up some obsolete - maybe even dead? code. Goodjob!

PR #55, opened 2021-06-11
DM-30711: Added new commands and modified telemetry of MTDome.

wvreeven:


tests/test_mtdome_csc.py
```diff
@@ -653,5 +691,87 @@ async def test_status_error(self):
...
+            initial_state=salobj.State.STANDBY,
+            config_dir=None,
+            simulation_mode=1,
+        ):
+            await self.set_csc_to_enabled()
+
+            # Prepare the lower level components
+            self.csc.mock_ctrl.amcs.status = MotionState.FAULT
+            self.csc.mock_ctrl.apscs.status = MotionState.FAULT
+            self.csc.mock_ctrl.lcs.status[:] = MotionState.FAULT.name
```
r-owen, line 705: Why is this line different? Is that difference necessary?

tests/test_mtdome_csc.py
```diff
@@ -653,5 +691,87 @@ async def test_status_error(self):
...
+            self.csc.mock_ctrl.lwscs.status = MotionState.FAULT
+            self.csc.mock_ctrl.moncs.status = MotionState.FAULT
+            self.csc.mock_ctrl.thcs.status = MotionState.FAULT
+            self.csc.mock_ctrl.amcs._commanded_motion_state = MotionState.FAULT
+            self.csc.mock_ctrl.lwscs._commanded_motion_state = MotionState.FAULT
+
+            await self.csc.write_then_read_reply(command="exitFault")
+
+            await self.csc.statusAMCS()
+            amcs_status = self.csc.lower_level_status[LlcName.AMCS.value]
```
r-owen, line 715: Is `.value` necessary here? I think the enum class is a subclass of enum.IntEnum so I would expect it not to be needed.

r-owen (changes requested):
This looks like a nice improvement. My main requested change is to improve the doc strings. In some cases they are incorrect (probably reflecting the initial implementation) and in others they are unclear.

I hope you will also consider trying to reduce repetition in several places, as per my in-line comments. If you agree but prefer to make such changes as a separate ticket, that's fine with me (though changing how you dispatch the stopX and goStationaryX commands in the mock controller would be a pretty simple change).

python/lsst/ts/MTDome/llc_motion_state.py
```diff
@@ -0,0 +1,42 @@
...
+    CLOSED = 1
+    CRAWLING = 2
+    MOVING = 3
+    OPEN = 4
+    PARKED = 5
+    PARKING = 6
+    STOPPED = 7
+    STOPPING = 8
+    # Used by the lower level components and need to be translated to
+    # IDL MotionState values.
```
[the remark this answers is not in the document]
wvreeven: The low-level controller uses strings so I don't think that will be possible.

python/lsst/ts/MTDome/mock_controller.py
```diff
@@ -503,6 +516,52 @@ async def park(self):
         self.log.info("Received command 'park'")
         return await self.amcs.park(self.current_tai)
 
+    async def go_stationary_az(self):
+        """Set the azimuth motion to stationary state.
```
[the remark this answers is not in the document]
wvreeven: I updated the doc strings of the `go_stationary_az` and `go_stationary_el` functions. I'll consider generating the methods in a separate ticket.

python/lsst/ts/MTDome/mock_llc/mock_motion/elevation_motion.py
```diff
@@ -88,16 +88,23 @@ def get_position_velocity_and_motion_state(self, tai):
...
             ]:
-                motion_state = MotionState.STOPPED
+                motion_state = LlcMotionState.STOPPED
+                position = self._end_position
+                velocity = 0
+            elif self._commanded_motion_state in [
+                LlcMotionState.GO_STATIONARY,
+                LlcMotionState.STATIONARY,
+            ]:
+                motion_state = LlcMotionState.STATIONARY
```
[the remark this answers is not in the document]
wvreeven: A very good suggestion but I'm afraid there is one major difference between the azimuth and elevation motion: the azimuth motion is supposed to start crawling after a move and the elevation motion is always supposed to stop. So I prefer to leave the code as is.

python/lsst/ts/MTDome/mtdome_csc.py
```diff
@@ -318,37 +319,49 @@ async def do_moveEl(self, data):
         self.evt_elTarget.set_put(position=data.position, velocity=0)
 
     async def do_stopAz(self, data):
-        """Stop AZ.
+        """Stop AZ, meaning stop all motion but do not engage the brakes nor
+        the locking pins.
```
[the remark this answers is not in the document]
wvreeven: You are right. This reflects the situation before I removed the `goStationary` functions again. Done.

tests/test_mtdome_csc.py
wvreeven: The statuses are passed on as strings, so it is necessary.

tests/test_mtdome_csc.py
wvreeven: The statuses are passed on as strings, so it is necessary.

python/lsst/ts/MTDome/mock_llc/amcs.py
```diff
@@ -203,11 +206,25 @@ async def park(self, start_tai):
...
             tests it can be convenient to use other values.
         """
-        self.status = MotionState.PARKING
+        self.status = LlcMotionState.PARKING
         self.position_commanded = PARK_POSITION
         self.duration = self.azimuth_motion.park(start_tai)
         return self.duration
 
+    async def go_stationary(self, start_tai):
+        """Set the dome to stationary state by stopping the motion.
```
r-owen, line 215: Brakes? Locking pins? The doc strings in the mock controller are very clear; please consider copying them to the mock-subsystem methods.

tests/test_mtdome_csc.py
r-owen: Ah, I see that LlcName is an enum.Enum, not an enum.IntEnum. If you made it inherit from str as well as enum.Enum then you could skip the `.value` but it's not a big deal.

r-owen (approved):
The doc strings for go_stationary and stop on the mock controller are great. I hope you will consider using them for the subsystem-specific versions of the commands as well.

python/lsst/ts/MTDome/mock_llc/amcs.py
wvreeven: Yes, I was being sloppy again. Done.

PR #57, opened 2021-08-12
DM-31391: Fixed the conda build. Also fixed some package export errors.

wvreeven:


tribeiro (approved):
Looks good! 2 comments though. 

1 - Do you want to update the version history?
2 - You could have separated the two changes (conda fix and Package exports(?)) in 2 different commits. And then one for the version history update.

PR #63, opened 2021-10-19
DM-31951: Added auto-enable capacity.

wvreeven:


mareuter (approved):
The work looks fine, but this is a feature we will probably never use for the MTDome. The auto-enable is usually used for components that do monitoring (ESS) or need to be available immediately when we do an upgrade (Watcher and ScriptQueues). The components that handle bigger hardware, we want folks to make a conscious decision to turn things on.

PR #65, opened 2021-11-22
Added 'appliedConfiguration' to the status replies of AMCS and LWSCS.

wvreeven:


couger01 (approved):
Except for one typo, looks good to me.

PR #67, opened 2022-01-04
DM-32671: Added new MotionStates to IDL.

wvreeven:


python/lsst/ts/mtdome/mtdome_csc.py
```diff
@@ -817,15 +785,6 @@ async def request_and_send_llc_status(
             else:
                 if llc_status["status"] == LlcMotionState.STATIONARY.name:
                     motion_state = MotionState.STOPPED_BRAKED
```
r-owen, line 787: Thanks for the simplification. Is it worth asking vendor to change "stationary" to "stopped_braked"? I agree it's a clearer term (since stationary does not imply the brakes are on).

r-owen (approved):
Nice simplification

python/lsst/ts/mtdome/mtdome_csc.py
wvreeven: Good suggestion. I'll ask in the next dome software meeting.

PR #68, opened 2022-01-26
DM-27932: Add network error handling.

wvreeven:


python/lsst/ts/mtdome/mtdome_csc.py
```diff
@@ -321,13 +324,21 @@ async def write_then_read_reply(
...
             self.log.debug(f"Sending command {st}")
-            assert self.writer is not None
-            self.writer.write(st.encode() + b"\r\n")
-            await self.writer.drain()
-            assert self.reader is not None
-            read_bytes = await asyncio.wait_for(
-                self.reader.readuntil(b"\r\n"), timeout=_TIMEOUT
-            )
+            try:
+                assert self.writer is not None
```
r-owen, line 328: Please consider checking that you are connected, both here and at line 335 (instead of `assert self.reader is not None`). It is a more useful test: there is no point reading or writing if the connection is gone.

python/lsst/ts/mtdome/mtdome_csc.py
```diff
@@ -321,13 +324,21 @@ async def write_then_read_reply(
...
+                self.writer.write(st.encode() + b"\r\n")
+                await self.writer.drain()
+            except (asyncio.exceptions.TimeoutError, AssertionError) as e:
```
r-owen, line 331: I'm curious: what triggers the timeout error? Does StreamWriter.drain raise this? Do you have any control over this? (Or is this just copied from the read branch and not actually needed)?

python/lsst/ts/mtdome/mtdome_csc.py
r-owen: In a similar vein, I see that your `connected` property only checks if the reader and writer are None. Please consider something a bit more detailed, such as (from ts_tcpip's OneClientServer):
```
    @property
    def connected(self) -> bool:
        """Return True if a client is connected to this server."""
        return not (
            self.reader is None
            or self.writer is None
            or self.reader.at_eof()
            or self.writer.is_closing()
        )
```

tests/test_mock_controller.py
```diff
@@ -67,15 +72,26 @@ async def asyncSetUp(self) -> None:
...
+            The timeout to use; default to 1 [s].
 
         Returns
         -------
         data : `dict`
             A dictionary with objects representing the string read.
+
+        Notes
+        -----
+        The timeout should only be set in case a network issue is mocked.
```
r-owen, line 90: I'm somewhat confused by this statement. What do you mean by a "network issue"?

I see that you use a longer value when simulating a slow network where you are willing to wait for the data. If that's what you have in mind, you could say that here, but:
* You have nice explanatory code comments where you do that below.
* It is a fairly self-evident use case.

So I suggest deleting the note. But if you are keen to keep it, then please clarify it.

tests/test_mock_controller.py
```diff
@@ -1182,3 +1198,26 @@ async def test_status(self) -> None:
         thcs_status = self.data[mtdome.LlcName.THCS.value]
         assert thcs_status["status"]["status"] == mtdome.LlcMotionState.CLOSED.name
         assert thcs_status["temperature"] == [0.0] * mtdome.mock_llc.NUM_THERMO_SENSORS
+
+    async def test_slow_network(self) -> None:
```
r-owen, line 1202: It took me some time to figure out the point of this test. Please consider a doc string such as:
```
"""Test the slow_network feature of the mock controller."""
```
 It's probably not a big help, but it might keep the user from wondering if you are just testing that "it takes longer to read data if you pretend the network is slow", which is not interesting.

tests/test_mock_controller.py
```diff
@@ -1182,3 +1198,26 @@ async def test_status(self) -> None:
...
+        with pytest.raises(asyncio.exceptions.TimeoutError):
+            # The default timeout should time out because of the slow network.
+            await self.read()
+        # Waiting longer should eventually result in a successful read.
+        self.data = await self.read(timeout=SLOW_NETWORK_TIMEOUT)
+        amcs_status = self.data[mtdome.LlcName.AMCS.value]
+        assert amcs_status["status"]["status"] == mtdome.LlcMotionState.PARKED.name
+        assert amcs_status["positionActual"] == 0
+
+    async def test_network_interruption(self) -> None:
```
r-owen, line 1214: Again, a brief doc string would be helpful.

tests/test_mtdome_csc.py
```diff
@@ -1067,5 +1067,59 @@ async def test_do_setOperationalMode(self) -> None:
...
+            await self.set_csc_to_enabled()
+            await self.assert_next_summary_state(salobj.State.ENABLED)
+            self.csc.mock_ctrl.network_interruption = True
+            await self.assert_next_summary_state(salobj.State.FAULT)
+
+    async def test_no_connection(self) -> None:
+        # Set simultation_mode deliberately to 0 to make sure that no mock
+        # server is started by the CSC. Also set the initial state to DISABLED
+        # so the RuntimeError, that gets raised when going to ENABLED, can be
+        # caught and processed.
```
r-owen, line 1110: Is this safe? I'd hate to think that this unit test might succeed in connecting to the real server, if run at the wrong location.

There are safer ways to simulate this, such as adding a flag to the mock controller that makes it refuse all connections (e.g. `enable_server`). It does take a bit of work, especially if you use port 0 for the server and then set it to an available port. But you can report the final port as 0; I think that will reliably make the client fail.

r-owen (approved):
Looks very nice. I do have some suggestions.

python/lsst/ts/mtdome/mock_controller.py
```diff
@@ -121,11 +124,17 @@ def __init__(
             "stopLouvers": self.stop_louvers,
             "stopShutter": self.stop_shutter,
         }
-        # Durations used by this class and by its unit test
+        # Durations used by this class and by its unit test.
```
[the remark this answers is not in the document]
wvreeven: I opted for class constants.

python/lsst/ts/mtdome/mock_controller.py
```diff
@@ -121,11 +124,17 @@ def __init__(
...
         self.long_duration = 20
         self.short_duration = 2
         # Time keeping
         self.current_tai = 0
+        # Mock a slow network (True) or not (False). To be set by unit tests
+        # only.
+        self.slow_network = False
+        # Mock a network interruption (True) or not (False). To be set by unit
+        # tests only.
+        self.network_interruption = False
```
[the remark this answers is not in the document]
wvreeven: Done

python/lsst/ts/mtdome/mtdome_csc.py
```diff
@@ -192,28 +192,33 @@ async def connect(self) -> None:
...
+            self.evt_lockingPinsEngaged.set_put(engaged=0)
+
+            # Start polling for the status of the lower level components
+            # periodically.
+            await self.start_status_tasks()
+
+            self.log.info("connected")
+        except ConnectionError as e:
+            self.fault(code=3, report=f"Connection to server failed: {e}")
+            raise
```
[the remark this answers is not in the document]
wvreeven: I moved the code out of the block.

python/lsst/ts/mtdome/mtdome_csc.py
```diff
@@ -291,8 +294,8 @@ async def handle_summary_state(self) -> None:
         disconnect to the lower level components (or the mock_controller) when
         needed.
         """
-        self.log.info(f"handle_summary_state {self.summary_state}")
-        if self.disabled_or_enabled:
+        self.log.info(f"handle_summary_state {self.summary_state.name}")
+        if self.summary_state == salobj.State.ENABLED:
```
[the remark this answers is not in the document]
wvreeven: You have a very good point and I reverted the change. It required a small modification to one of the unit test cases as well since the CSC will go to FAULT state already when going to DISABLED instead of ENABLED because the status commands time out as well as all other commands.

python/lsst/ts/mtdome/mtdome_csc.py
wvreeven: I tried that a long time ago but then MyPy isn't happy:

`
329: error: Item "None" of "Optional[StreamWriter]" has no attribute "write"
330: error: Item "None" of "Optional[StreamWriter]" has no attribute "drain"
337: error: Item "None" of "Optional[StreamReader]" has no attribute "readuntil"
`

tests/test_mock_controller.py
wvreeven: I removed the note.

tests/test_mock_controller.py
wvreeven: Doc string added.

tests/test_mock_controller.py
wvreeven: Done

tests/test_mtdome_csc.py
wvreeven: That's a very good point that I didn't consider at all! I have introduced a variable in the CSC that can be passed on to the mock controller which, when set to False, will stop the mock controller immediately after starting it. That way, a port gets allocated and the CSC code doesn't break.

python/lsst/ts/mtdome/mtdome_csc.py
wvreeven: I copied from the read branch and tried without catching the timeout error. The unit tests pass so it indeed is not needed. Thanks for spotting that!

python/lsst/ts/mtdome/mtdome_csc.py
r-owen: Yes. I know. I use `# type: ignore` to get around that. It's worth it to me to a more complete "is connected" test. But if you really hate `# type: ignore` then you can also assert that the thing is not None. It's just that the full `connected` test is more useful in reality. The other just is a sop to mypy.

PR #69, opened 2022-03-04
CAP-856: Prepare for salobj 7.

wvreeven:


tests/test_mtdome_csc.py
```diff
@@ -64,7 +64,7 @@ async def test_standard_state_transitions(self) -> None:
             initial_state=salobj.State.STANDBY,
             config_dir=None,
             simulation_mode=1,
-            settings_to_apply="",
+            override="",
```
r-owen, line 67: This is the default. You might consider omitting it.

r-owen: I am quite concerned about config_dir=None. I think you should change these all to point to `tests/data/config` so that you can be sure exactly what config will be used.

The risk of using the default config dir (ts_config_mttcs) is that if some of the config is site-specific (likely so) then `_init.yaml` isn't complete and if there is no `_test.yaml` your tests will fail due to incomplete configuration.

Another option is to add `_test.yaml` to `ts_config_mttcs` as long as you don't mind having somewhat harder-to-control config. But I feel result is more fragile than having the full config in your own package.

r-owen (commented):
Please let me know if you decide to add local config and ask me to review it again.

tests/test_mtdome_csc.py
wvreeven: Both done. Good suggestion to add a test `_init.yaml` config file, thanks!

tests/test_mtdome_csc.py
```diff
@@ -48,23 +48,20 @@ def basic_make_csc(
...
             simulation_mode=simulation_mode,
-            settings_to_apply=settings_to_apply,
             mock_port=0,
         )
 
     async def test_standard_state_transitions(self) -> None:
         async with self.make_csc(
             initial_state=salobj.State.STANDBY,
-            config_dir=None,
+            config_dir="tests/data/config",
```
r-owen, line 63: I'm surprised this works and fear it is unsafe -- for instance it may require that tests only be run within the test directory (`pytest tests/test_csc.py` would fail but `cd tests; pytest test_csc.py` would pass). I suggest putting this before the test class:
```
CONFIG_DIR = pathlib.Path(__file__).parent / "data" / "config"
```
and then using `config_dir=CONFIG_DIR` everywhere.

r-owen (approved):
Thank you for the updates. I have one easy and likely important requested change: a more robust way of specifying the config dir in your unit tests.

tests/test_mtdome_csc.py
wvreeven: Oops quite a beginner's mistake. Thanks for educating me once more.

PR #71, opened 2022-05-04
Tickets/dm 34524

wvreeven:


python/lsst/ts/mtdome/mtdome_csc.py
```diff
@@ -40,26 +41,56 @@
...
     "velocityCommanded",
-}
+}.union(_AMCS_KEYS_OFFSET)
+
+# The offset of the dome rotation zero point with respect to azimuth 0º (true
+# north) is 32º west and the aperture lies at the opposite side of the dome.
+# Therefore the dome azimuth offset is 180º - 32º = 148º and this needs to be
+# subtracted when commanding the azimuth position, or added when sending the
+# azimuth telemetry.
+DOME_AZIMUTH_OFFSET = 148.0
```
r-owen, line 72: How stable is this likely to be? If it may change or require refinement, please consider making it a configuration parameter.

python/lsst/ts/mtdome/mtdome_csc.py
```diff
@@ -40,26 +41,56 @@
...
 _LCS_STATUS_PERIOD = 2.0
 _LWSCS_STATUS_PERIOD = 2.0
 _MONCS_STATUS_PERIOD = 2.0
 _THCS_STATUS_PERIOD = 2.0
 
+# These next commands are temporarily disabled because they will be issued
+# during the upcoming TMA pointing test and the EIE LabVIEW code doesn't handle
+# them yet, which will result in an error. As soon as the TMA pointing test is
+# done, they will be reenabled.
+COMMANDS_DISABLED_FOR_TMA_POINTING = {
```
r-owen, line 86: I suspect you might want to change this set over time as we perform additional commissioning tests. If you agree, please consider a name such as DISABLED_COMMANDS (relying on the code comment to explain why) or perhaps something like COMMANDS_DISABLED_FOR_COMMISSIONING.

python/lsst/ts/mtdome/mtdome_csc.py
```diff
@@ -40,26 +41,56 @@
...
+    "crawlEl",
+    "goStationaryEl",
+    "moveEl",
+    "stopEl",
+}
+DATA_FOR_TMA_POINING = {"response": 0, "timeout": 0}
```
r-owen, line 92: I gently suggest a name like REPLY_DATA_FOR_DISABLED_COMMANDS. It took me awhile to figure out what it was.

If you keep the name, I suspect POINING should be POINTING.

python/lsst/ts/mtdome/mtdome_csc.py
```diff
@@ -334,21 +365,28 @@ async def write_then_read_reply(
...
+                data: typing.Dict[str, typing.Any] = encoding_tools.decode(
+                    read_bytes.decode()
                 )
-            except (asyncio.exceptions.TimeoutError, AssertionError) as e:
-                await self.fault(
-                    code=3, report=f"Error reading reply to command {st}: {e}"
+            else:
+                data = DATA_FOR_TMA_POINING
+                self.log.debug(
+                    f"WOUTERRRR command {command!r} resulted in reply {data}"
```
r-owen, line 388: I'm thinking that perhaps this should be a warning? Though if these commands are sent often then that could easily get annoying and debug is better.

I gently suggest "DISABLED command..."

tests/test_mtdome_csc.py
```diff
@@ -536,15 +539,8 @@ async def test_do_park(self) -> None:
...
-            # Now also check the azMotion event.
-            await self.csc.statusAMCS()
-            amcs_status = self.csc.lower_level_status[mtdome.LlcName.AMCS.value]
-            assert amcs_status["status"]["status"] == MotionState.PARKED.name
-            await self.assert_next_sample(
-                topic=self.remote.evt_azMotion,
-                state=MotionState.PARKED,
-                inPosition=True,
-            )
+            # No new azMotion event gets emitted since the PARKED event already
```
r-owen, line 542: I think the code at line 537 is no longer relevant. you change mock_ctrl.current_tai but don't do anything to test for the effect of that.

If you do want to test for state I suppose you could still check the state of the CSC. Checking events is trickier and may well not be worth the hassle. I sometimes read until I get the value I'm looking for (which times out if the value never comes). But it can be hard to be sure that the match you find is for the most recent message.

r-owen (approved):
This looks good. My only substantial concern is hard-coding the offset. If there are reasons to expect the value you provide to be exact and stable, then please explain that in a code comment. Otherwise please make it a config parameter.

It is a great pity that the zero point of the dome cannot be set to match the telescope.

python/lsst/ts/mtdome/mtdome_csc.py
wvreeven: This is 100% stable since it depends on the fixed position of the hardware locking pins. The only thing that may change is a refinement of the value. It could become 147.87 or 148.46 but otherwise this is fixed. I'll leave it hard coded.

python/lsst/ts/mtdome/mtdome_csc.py
wvreeven: Good suggestion. Done.

python/lsst/ts/mtdome/mtdome_csc.py
wvreeven: Good suggestion. I renamed it.

python/lsst/ts/mtdome/mtdome_csc.py
wvreeven: Thanks for catching this. I forgot to remove this log statement since I used it for debugging the unit tests. I removed it now.

tests/test_mtdome_csc.py
wvreeven: Good catch. I removed the lines.

PR #80, opened 2022-08-17
Refactor the BaseLlcMotion class to have subclasses with and without …

wvreeven:
…crawl.

### DIFF
--- a/python/lsst/ts/mtdome/mock_llc/mock_motion/base_llc_motion.py
+++ b/python/lsst/ts/mtdome/mock_llc/mock_motion/base_llc_motion.py
@@ -48,6 +48,8 @@ class BaseLlcMotion(ABC):
         # This is not a constant but can be configured by the MTDome CSC, which
         # is why it is a parameter.
         self._max_speed = max_speed
+        # Keep track of the current motion state.
+        self._current_motion_state = LlcMotionState.PARKED
         # The commanded LlcMotionState, against which the computed
         # LlcMotionState will be compared. By default the elevation motion
         # starts in STOPPED state. The LlcMotionState only changes when a new
@@ -62,9 +64,69 @@ class BaseLlcMotion(ABC):
         # command. To model the real dome, this should be the current time.
         # However, for unit tests it can be convenient to use other values.
         self._end_tai = 0.0
-        # When a move or crawl command is received, it specifies the crawl
-        # crawl_velocity.
-        self._crawl_velocity = 0.0
+        # Keep track of being in error state or not.
+        self.motion_state_in_error = False
+        # Keep track of which drives are in error state. For this base class
+        # this is just a dummy value to make the code work. Each implementing
+        # subclass will have to define this properly itself.
+        self.drives_in_error_state = [False]
+        # Keep track of when motion state went into ERROR
+        self._error_start_tai = 0.0
+        # Keep track of the position at the moment that ERROR state starts.
+        self._error_state_position = 0.0
+
+    def base_set_target_position_and_velocity(
+        self,
+        start_tai: float,
+        end_position: float,
+        motion_state: LlcMotionState,
+    ) -> float:
+        """Sets the end_position and crawl_velocity and returns the duration of
+        the move.
+
+        No aceleration is taken into account.
+
+        Parameters
+        ----------
+        start_tai: `float`
+            The TAI time, unix seconds, at which the command was issued. To
+            model the real dome, this should be the current time. However, for
+            unit tests it can be convenient to use other values.
+        end_position: `float`
+            The target position.
+        motion_state: `LlcMotionState`
+            MOVING or CRAWLING. The value is not checked.
+
+        Returns
+        -------
+        duration: `float`
+            The duration of the move.
+
+        Raises
+        ------
+        ValueError
+            If the target position falls outside the range
+            [min position, max position] or if abs(crawl_velocity) > max_speed.
+
+        """
+        if motion_state not in [LlcMotionState.MOVING, LlcMotionState.CRAWLING]:
+            raise ValueError(f"{motion_state=!r} should be MOVING or CRAWLING.")
+
+        if (
+            motion_state == LlcMotionState.MOVING
+            and not self._min_position <= end_position <= self._max_position
+        ):
+            raise ValueError(
+                f"The target position {end_position} is outside of the "
+                f"range [{self._min_position, self._max_position}]"
+            )
+
+        self._commanded_motion_state = motion_state
+        self._start_tai = start_tai
+        self._end_position = end_position
+        duration = self._get_duration()
+        self._end_tai = self._start_tai + duration
+        return duration
 
     def _get_distance(self) -> float:
         """Determines the smallest distance [rad] between the initial and
@@ -74,6 +136,11 @@ class BaseLlcMotion(ABC):
         -------
         distance: `float`
             The smallest distance between the initial and target positions.
+
+        Notes
+        -----
+        If "_start_position" and "_end_position" are not in radians, this
+        method should be overridden.
         """
         distance = utils.angle_diff(
             math.degrees(self._end_position), math.degrees(self._start_position)
@@ -90,22 +157,16 @@ class BaseLlcMotion(ABC):
             The duration of the move, or zero in case of a crawl.
         """
         duration = math.fabs(self._get_distance()) / self._max_speed
-        if self._commanded_motion_state == LlcMotionState.CRAWLING:
-            # A crawl command is executed instantaneously.
-            duration = 0
         return duration
 
-    def set_target_position_and_velocity(
-        self,
-        start_tai: float,
-        end_position: float,
-        crawl_velocity: float,
-        motion_state: LlcMotionState,
-    ) -> float:
-        """Sets the end_position and crawl_velocity and returns the duration of
-        the move.
+    @abstractmethod
+    def get_position_velocity_and_motion_state(
+        self, tai: float
+    ) -> tuple[float, float, LlcMotionState]:
+        pass
 
-        No aceleration is taken into account.
+    def stop(self, start_tai: float) -> float:
+        """Stop the current motion.
 
         Parameters
         ----------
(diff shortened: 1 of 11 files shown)


python/lsst/ts/mtdome/mock_llc/mock_motion/base_llc_motion_without_crawl.py
```diff
@@ -0,0 +1,55 @@
...
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+from abc import ABC
+
+from ...enums import LlcMotionState
+from .base_llc_motion import BaseLlcMotion
+
+
+class BaseLlcMotionWithoutCrawl(BaseLlcMotion, ABC):
```
r-owen, line 28: Why the ABC here? Is it because get_position_velocity_and_motion_state is still not defined? Is it necessary?

r-owen: In a related question: would it practical to define get_position_velocity_and_motion_state in the two BaseLlcMotionWith/WithoutCrawl classes, rather than leaving that for more detailed subclasses?

I'm surprised how different the three implementations are, so perhaps not. But I figured I'd ask. I confess some of them are so complicated that it makes me rather nervous -- it looks hard to test all possible cases.

python/lsst/ts/mtdome/mock_llc/mock_motion/base_llc_motion_without_crawl.py
```diff
@@ -0,0 +1,55 @@
...
+        max_speed: float,
+        start_tai: float,
+    ):
+        super().__init__(
+            start_position=start_position,
+            min_position=min_position,
+            max_position=max_position,
+            max_speed=max_speed,
+            start_tai=start_tai,
+        )
```
r-owen, line 43: Please delete this method because it just duplicates the base class constructor.

python/lsst/ts/mtdome/mock_llc/mock_motion/shutter_motion.py
```diff
@@ -72,43 +72,15 @@ def __init__(
...
-        # Keep track of being in error state or not.
-        self.motion_state_in_error = False
         # Keep track of which drives are in error state.
         self.drives_in_error_state = [False] * NUM_MOTORS_PER_SHUTTER
-        # Keep track of when motion state went into ERROR
-        self._error_start_tai = 0.0
-        # Keep track of the position at the moment that ERROR state starts.
-        self._error_state_position = 0.0
 
     def _get_duration(self) -> float:
```
r-owen, line 78: Is this method still needed? I hope you can delete it (short though it is).

python/lsst/ts/mtdome/mock_llc/mock_motion/elevation_motion.py
```diff
@@ -157,35 +157,20 @@ def get_position_velocity_and_motion_state(
         position = utils.angle_wrap_nonnegative(math.degrees(position)).rad
         return position, velocity, motion_state
 
-    def stop(self, start_tai: float) -> float:
-        """Stops the current motion.
+    def exit_fault(self, start_tai: float) -> float:
+        """Clear the fault state.
```
r-owen, line 161: Is this still needed? I hope you can use the version in the parent class, but if not, please explain why in the doc string.

python/lsst/ts/mtdome/mock_llc/mock_motion/elevation_motion.py
```diff
@@ -197,21 +182,5 @@ def go_stationary(self, start_tai: float) -> float:
...
-        position, velocity, motion_state = self.get_position_velocity_and_motion_state(
-            tai=start_tai
-        )
-        self._start_tai = start_tai
-        self._start_position = position
-        self._end_position = position
-        self._crawl_velocity = 0.0
-        self._commanded_motion_state = LlcMotionState.STATIONARY
+    def set_fault(self, start_tai: float, drives_in_error: list[int]) -> None:
+        raise NotImplementedError
```
r-owen, line 186: Please document why this is not supported.

python/lsst/ts/mtdome/mock_llc/mock_motion/base_llc_motion_with_crawl.py
```diff
@@ -0,0 +1,173 @@
...
+        Returns
+        -------
+        duration: `float`
+            The duration of the move, or zero in case of a crawl.
+        """
+        duration = super()._get_duration()
+        if self._commanded_motion_state == LlcMotionState.CRAWLING:
+            # A crawl command is executed instantaneously.
+            duration = 0
+        return duration
```
r-owen, line 63: Why call super()._get_duration() before testing if crawling? Does it change internal state or do some error checking? Please consider adding a code comment explaining, since the naive thing to do is only call super()._get_duration() if not crawling, and readers will want to know.

r-owen (approved):
Looks great! I have a few minor suggestions, and a question about whether the get_position_velocity_and_motion_state implementations can be simplified a bit.

python/lsst/ts/mtdome/mock_llc/mock_motion/base_llc_motion.py
```diff
@@ -48,6 +48,8 @@ def __init__(
         # This is not a constant but can be configured by the MTDome CSC, which
         # is why it is a parameter.
         self._max_speed = max_speed
+        # Keep track of the motion state where we came from.
```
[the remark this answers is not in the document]
wvreeven: It isn't necessarily the most recently commanded motion state, since this may be updated internally. So it is the last known motion state or simply the current motion state. I have updated the comment.

python/lsst/ts/mtdome/mock_llc/mock_motion/base_llc_motion.py
```diff
@@ -62,44 +64,21 @@ def __init__(
         # command. To model the real dome, this should be the current time.
         # However, for unit tests it can be convenient to use other values.
         self._end_tai = 0.0
-        # When a move or crawl command is received, it specifies the crawl
-        # crawl_velocity.
-        self._crawl_velocity = 0.0
+        # Keep track of being in error state or not.
+        self.motion_state_in_error = False
```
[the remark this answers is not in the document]
wvreeven: I have considered computing it from `self._error_start_tai` but that would complicate the code too much. The low level components have an error state that is independent of the motor errors so `self.drives_in_error_state` cannot be used. I have kept this as is.

python/lsst/ts/mtdome/mock_llc/mock_motion/base_llc_motion.py
```diff
@@ -132,25 +109,184 @@ def set_target_position_and_velocity(
             [min position, max position] or if abs(crawl_velocity) > max_speed.
 
         """
-        if not self._min_position <= end_position <= self._max_position:
+        if motion_state not in [LlcMotionState.MOVING, LlcMotionState.CRAWLING]:
+            raise ValueError("motion_speed should be MOVING or CRAWLING.")
```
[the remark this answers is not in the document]
wvreeven: Thanks, I didn't know about the `=!r` construction. Done.

python/lsst/ts/mtdome/mock_llc/mock_motion/base_llc_motion.py
```diff
@@ -132,25 +109,184 @@ def set_target_position_and_velocity(
...
+
+        if (
+            motion_state == LlcMotionState.MOVING
+            and not self._min_position <= end_position <= self._max_position
+        ):
```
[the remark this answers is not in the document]
wvreeven: Because when crawling it could be crawling over the 360º limit so it needs to be set to some very high value so crawling doesn't stop.

python/lsst/ts/mtdome/mock_llc/mock_motion/base_llc_motion.py
```diff
@@ -132,25 +109,184 @@ def set_target_position_and_velocity(
...
         self._start_tai = start_tai
         self._end_position = end_position
-        self._crawl_velocity = crawl_velocity
         duration = self._get_duration()
         self._end_tai = self._start_tai + duration
         return duration
 
+    def _get_distance(self) -> float:
+        """Determines the smallest distance [rad] between the initial and
+        target positions assuming motion around a circle.
```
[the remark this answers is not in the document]
wvreeven: It does also work for elevation since that uses angles as well. But it will not work for the aperture shutter since that is a linear motion and doesn't use angles so the conversion to degrees would mess up the result. I added a note that systems that do not use radians should override the method.

python/lsst/ts/mtdome/mock_llc/mock_motion/base_llc_motion_with_crawl.py
```diff
@@ -0,0 +1,173 @@
...
+        self._crawl_velocity = crawl_velocity
+        duration = self.base_set_target_position_and_velocity(
+            start_tai=start_tai,
+            end_position=end_position,
+            motion_state=motion_state,
+        )
+        return duration
+
+    def stop(self, start_tai: float) -> float:
+        """Stops the current.
```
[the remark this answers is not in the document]
wvreeven: Thanks for catching this. Done.

python/lsst/ts/mtdome/mock_llc/mock_motion/base_llc_motion_with_crawl.py
```diff
@@ -0,0 +1,173 @@
...
+
+
+        Returns
+        -------
+        `float`
+            The expected duration of the command.
+        """
+        super().stop(start_tai=start_tai)
+        self._end_tai = self._start_tai + self._get_duration()
+        return self._end_tai - start_tai
```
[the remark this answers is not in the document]
wvreeven: Good point. I moved those lines, as well from go_stationary, to the base class and removed both methods.

python/lsst/ts/mtdome/mock_llc/mock_motion/base_llc_motion_without_crawl.py
wvreeven: `ABC` is there indeed because of `get_position_velocity_and_motion_state`. When I remove `ABC`, PyCharm complains but flake8 and mypy don't so I removed it and the import.

wvreeven: Regarding your related question: the methods are quite different for azimuth and elevation. Later on the louvers will be added as well and those will have a very different implementation than the shutter. So I'll leave this as is.

python/lsst/ts/mtdome/mock_llc/mock_motion/base_llc_motion_without_crawl.py
wvreeven: Done

python/lsst/ts/mtdome/mock_llc/mock_motion/shutter_motion.py
wvreeven: Yes it is. This class doesn't use angles (in radians) but a percentage. It cannot use the superclass method. I have added a comment explaining that.

python/lsst/ts/mtdome/mock_llc/mock_motion/elevation_motion.py
wvreeven: Done.

python/lsst/ts/mtdome/mock_llc/mock_motion/base_llc_motion_with_crawl.py
wvreeven: The super method computes the duration for the default case, which is MOVING. This class extends that for CRAWLING. I have added comments to explain that.

python/lsst/ts/mtdome/mock_llc/mock_motion/elevation_motion.py
wvreeven: I added a comment explaining why it is still needed.

PR #83, opened 2022-09-08
Tickets/dm 36122

wvreeven:


doc/version_history.rst
```diff
@@ -25,6 +25,7 @@ Requires:
 v1.11.3
 =======
 * Refactor the other test cases to contain less duplicate code.
+* Rename the 'searchZeroShutter' command to 'home' and make it apply to all subsystems.
```
tribeiro, line 28: uhm.. not sure this change will classify as a patch.. consider making this v1.12.0

tribeiro (approved):
Looks good, I would just recommend updating the version number..

doc/version_history.rst
wvreeven: Done

PR #84, opened 2022-09-14
u/wvreeven/tma pointing

wvreeven:


python/lsst/ts/mtdome/mtdome_csc.py
```diff
@@ -109,6 +109,23 @@ def run_mtdome() -> None:
     asyncio.run(MTDomeCsc.amain(index=None))
 
 
+def support_dds_item(command_name: str) -> bool:
```
tribeiro, line 112: Consider making a utils module for this.

python/lsst/ts/mtdome/mtdome_csc.py
```diff
@@ -261,7 +289,10 @@ async def connect(self) -> None:
         # DM-35794: Also send enabled event for Aperture Shutter.
         await self.evt_azEnabled.set_write(state=EnabledState.ENABLED)
         await self.evt_elEnabled.set_write(state=EnabledState.ENABLED)
-        await self.evt_shutterEnabled.set_write(state=EnabledState.ENABLED)
+        # Check supported event to make sure of backward compatibility with
+        # XML 12.0.
+        if support_dds_item("evt_shutterEnabled"):
```
tribeiro, line 294: I think this method is only useful for commands because you need to make the introspection before calling `init` of the parent class.. but at this point you could simply do:

```
if hasattr(self, "evt_shutterEnabled"):
```

which won't require you to parse the idl file.

python/lsst/ts/mtdome/mtdome_csc.py
```diff
@@ -909,9 +981,12 @@ async def _check_errors_and_send_events_shutter(
             fault_code = ", ".join(
                 [f"{message['code']}={message['description']}" for message in messages]
             )
-            await self.evt_shutterEnabled.set_write(
-                state=EnabledState.FAULT, faultCode=fault_code
-            )
+            # Check supported event to make sure of backward compatibility with
+            # XML 12.0.
+            if support_dds_item("evt_shutterEnabled"):
```
tribeiro, line 986: Again, use `hasattr` to avoid parsing the idl file.

tribeiro (approved):
Looks good, I did left a few inline comments I hope you will consider before merging.

python/lsst/ts/mtdome/mtdome_csc.py
wvreeven: Done

python/lsst/ts/mtdome/mtdome_csc.py
wvreeven: Good tip, thanks! Done.

python/lsst/ts/mtdome/mtdome_csc.py
wvreeven: Done

PR #85, opened 2022-10-03
Tickets/dm 36401

wvreeven:


python/lsst/ts/mtdome/enums.py
```diff
@@ -20,56 +20,26 @@
...
-    MOVING = enum.auto()
-    OPEN = enum.auto()
-    PARKED = enum.auto()
+class InternalMotionState(enum.IntEnum):
+    """Internal Motion states.
+
+    These get translated into IDL MotionState instances by the CSC.
+    """
+
+    DISABLED = enum.auto()
```
r-owen, line 41: I am quite uneasy about this enum because it has no translation information and because it sounds like an enum of internal motion states (which might be a very useful thing to have for your mock code, if it was complete).

Please consider saving the translation info as a dict of internal state name: external state name. That way the translation is obvious and also it is clear why it's short.

Please also consider programmatically generating a full InternalMotionState enum class (or using the old name LlcMotionState) for use in your mock code. This should be fairly easy to generate from the MotionState enum and the translation dict (probably the inverse of this dict, which is also easy to make on the fly). In case you are not familiar with the enum functional API, here is a trivial example (which happens to have values != names, but in your case I think you want them to match):
```
name_value_list = (("ANAME", "AVALUE"), ("BNAME", "BVALUE"))
Example = enum.Enum("Example", name_value_list)
```

r-owen (approved):
This works, but I hope you will be able to clarify the translation code and consider using a full internal motion state enum class in your mocks.

python/lsst/ts/mtdome/enums.py
wvreeven: I added a `motion_state_translations` dict that contains the translation info that now is used by the CSC. This has simplified the code there. Thanks for the suggestion.

As for the programmatically generated enum, this works in the code but MyPy goes completely berserk over it. I prefer leaving the code as is over adding `# type: ignore` comments everywhere.

PR #94, opened 2023-01-26
Enable the possibility to set the maximum velocity, acceleration and …

wvreeven:
…jerk for the azimuth rotation via the configuration.


couger01 (approved):
LGTM.

PR #97, opened 2023-03-31
Tickets/dm 38299

wvreeven:


python/lsst/ts/mtdome/mtdome_csc.py
```diff
@@ -530,6 +540,45 @@ async def write_then_read_reply(
...
+        because, if the velocity != 0.0, the dome is following a moving target
+        and the chance of it being at exactly the commanded position with the
+        commanded velocity can be considered zero and therefore the moveAz
+        command has to be sent to the dome.
+        """
+        if (
+            math.isclose(velocity, 0.0)
+            and math.isclose(position, self.latest_moveAz_command["position"])
+            and math.isclose(velocity, self.latest_moveAz_command["velocity"])
+        ):
```
r-owen, line 576: I strongly suggest making abs_tol explicit (I think abs_tol is more meaningful than rel_tol for this particular test) -- either as a pair of module constants (one for position, one for velocity) or a pair of arguments to the method with suitable defaults.

Otherwise there's no obvious way to tune the code to work with whatever limitations the dome has, and readers have to look up the function in the Python docs to see what's going on.

It may suffice to use the same abs_tol for both position and velocity, but by keeping them separate you can more easily tune them and it makes the units clear.

python/lsst/ts/mtdome/mtdome_csc.py
```diff
@@ -279,6 +279,16 @@ def __init__(
...
 
+        # Keep track of the parameters of the latest moveAz command issued so
+        # repetition of the same command can be avoided. This is necessary in
+        # case the dome repeatedly is instructed to move to the same position
+        # with velocity == 0.0 since that may introduce large oscillations. The
+        # position is set to the dome park position.
+        self.latest_moveAz_command = {
+            "position": 360.0 - DOME_AZIMUTH_OFFSET,
+            "velocity": 0.0,
+        }
```
r-owen, line 290: Please consider a dataclass or an instance of the move command topic data. That reduces the risk of mis-spelling field names.

Saving topic data is a bit simpler because there is no need to define the data class, and you can literally just say `self.latst_moveAz_command = data` in do_moveAz. Also it provides a timestamp, in case it's of interest. But it does add some irrelevant fields. If you use the topic data then you can initialize it in the `__init__` after calling `super().__init__` with `self.latst_moveAz_command = self.cmd_moveAz.DataType(position=..., velocity=...)`, which avoids the need to test for null. You can recognize "never set" by testing that private_sndStamp=0.

python/lsst/ts/mtdome/mtdome_csc.py
```diff
@@ -279,6 +279,16 @@ def __init__(
         #  by the AMCS.
         self.previous_state = MotionState.PARKED
 
+        # Keep track of the parameters of the latest moveAz command issued so
+        # repetition of the same command can be avoided. This is necessary in
+        # case the dome repeatedly is instructed to move to the same position
+        # with velocity == 0.0 since that may introduce large oscillations. The
+        # position is set to the dome park position.
+        self.latest_moveAz_command = {
+            "position": 360.0 - DOME_AZIMUTH_OFFSET,
```
r-owen, line 288: Please consider initializing the position to NaN. That way you will always get "no match" for the first moveAz command, and any reader can instantly see that.

r-owen (changes requested):
In addition to the specific comments, I have a two more general questions/comments:
* What happens if the CSC is enabled and the very first move matches what the dome is already doing? If that causes oscillation then you should update the saved value based on telemetry (at least while in disabled state, though doing it all the time also works). if you do that, then you should prevent going to enabled state until this telemetry has been seen. Unfortunately using telemetry affects how you save the data -- my simple suggestion of saving the last command data doesn't work very well.
* Please reset the saved value every time you go out of a state where it can be set. If you never use telemetry to set it, then reset it when not in enabled state. If you use telemetry then reset it when you disconnect from, or connect to, the low-level controller.

python/lsst/ts/mtdome/mtdome_csc.py
r-owen: Also, as noted in general comments: I suspect you should initialize this when you go out of enabled state, or when disconnected from the low-level controller (which one depends on if you also set it based on telemetry, and in which states you do that).

python/lsst/ts/mtdome/mtdome_csc.py
wvreeven: I opted for a dataclass.

python/lsst/ts/mtdome/mtdome_csc.py
wvreeven: Both done. I opted not to use telemetry since issuing a single moveAz command will not cause vibrations and the hassle of making all the required changes doesn't outweigh handling that case. The `latest_moveAz_command` now is reset when disconnecting from the low-level controller.

python/lsst/ts/mtdome/mtdome_csc.py
wvreeven: Done

python/lsst/ts/mtdome/mtdome_csc.py
```diff
@@ -167,6 +170,14 @@ def run_mtdome() -> None:
     asyncio.run(MTDomeCsc.amain(index=None))
 
 
+@dataclass
+class MoveAzCommandData:
+    """Class representing the data for a moveAz command."""
```
r-owen, line 175: Please document the units. This is especially important for the dome because it uses degrees in some places and radians in others.

python/lsst/ts/mtdome/mtdome_csc.py
```diff
@@ -439,6 +456,7 @@ async def handle_summary_state(self) -> None:
                 await self.connect()
         else:
             await self.disconnect()
+            self.latest_moveAz_command = MoveAzCommandData()
```
r-owen, line 459: What happens if disconnect fails? Does the CSC go to FAULT or remain connected and enabled or...? That affects whether this line should appear before disconnecting or after. 

I simplest/clearest option may be to move this line to right before self.connect() above. Then it should be reliably run, regardless of what happens if disconnect fails.

tests/test_mtdome_csc.py
```diff
@@ -116,6 +116,29 @@ async def test_version(self) -> None:
...
+                position=360.0 - mtdome.DOME_AZIMUTH_OFFSET, velocity=0.0
+            )
+            for position, velocity in [
+                (360.0 - mtdome.DOME_AZIMUTH_OFFSET, 0.1),
+                (0.0, 0.0),
+                (0.0, 0.1),
+            ]:
+                assert not self.csc.assert_moveAz_command_is_same_as_latest(
+                    position=position, velocity=velocity
+                )
```
r-owen, line 140: Please add a few tests based on your tolerances.

r-owen (approved):
Looks good. I have a few minor suggestions I hope you will consider before merging.

python/lsst/ts/mtdome/mtdome_csc.py
```diff
@@ -530,6 +548,59 @@ async def write_then_read_reply(
 
             return data
 
+    def assert_moveAz_command_is_same_as_latest(
+        self, position: float, velocity: float
+    ) -> bool:
+        """Assert that the received moveAz command is the same as the latest
```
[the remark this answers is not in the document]
wvreeven: I opted for `is_moveAz_same_as_current` and renamed the `latest_moveAz_command` instance variable to `current_moveAz_command`.

PR #98, opened 2023-05-17
Tickets/dm 39129

wvreeven:


python/lsst/ts/mtdome/mock_controller.py
```diff
@@ -794,12 +768,13 @@ async def search_zero_shutter(self) -> float:
...
-    logging.info("main method")
+    log = logging.getLogger("MockMTDomeController")
+    log.info("main method")
     # An arbitrarily chosen port. Nothing special about it.
     port = 5000
-    logging.info("Constructing mock controller.")
-    mock_ctrl = MockMTDomeController(port=port)
-    logging.info("Starting mock MTDome controller.")
+    log.info("Constructing mock controller.")
+    mock_ctrl = MockMTDomeController(host=tcpip.DEFAULT_LOCALHOST, port=port, log=log)
```
r-owen, line 776: I gently suggest making MockMTDomeController have a default host of tcpip.DEFAULT_LOCALHOST so you don't have to specify it here.

You could even eliminate the host constructor argument in MockMTDomeController (I would probably do that).

python/lsst/ts/mtdome/mock_llc/mock_motion/base_llc_motion_with_crawl.py
```diff
@@ -19,14 +19,15 @@
...
+import abc
 import math
 
 from lsst.ts.idl.enums.MTDome import MotionState
 
 from .base_llc_motion import BaseLlcMotion
 
 
-class BaseLlcMotionWithCrawl(BaseLlcMotion):
+class BaseLlcMotionWithCrawl(BaseLlcMotion, abc.ABC):
```
r-owen, line 30: Why this change? I thought abstraction was inherited.

python/lsst/ts/mtdome/mtdome_csc.py
```diff
@@ -53,8 +53,7 @@
 from .llc_configuration_limits import AmcsLimits, LwscsLimits
 from .mock_controller import MockMTDomeController
 
-_LOCAL_HOST = "127.0.0.1"
-_TIMEOUT = 20  # timeout [sec] to be used by this module
+_TIMEOUT = 20  # timeout [sec] to be used by this module.
```
r-owen, line 56: Timeout of what? A bit more info would be helpful.

python/lsst/ts/mtdome/mtdome_csc.py
```diff
@@ -122,13 +125,15 @@
...
     ("statusThCS", _THCS_STATUS_PERIOD),
+    ("check_all_commands_have_replies", _COMMANDS_REPLIED_PERIOD),
 }
 
 # The status methods and the intervals at which they are executed to be used
 # during commissioning. Not all can be used because the LabVIEW code of EIE
 # doesn't support all of them yet.
 METHODS_AND_INTERVALS_FOR_COMMISSIONING = {
     ("statusAMCS", _AMCS_STATUS_PERIOD),
+    ("check_all_commands_have_replies", _COMMANDS_REPLIED_PERIOD),
```
r-owen, line 136: Please consider eliminating `METHODS_AND_INTERVALS_FOR_COMMISSIONING` and adding one more item to the entries in ALL_METHODS_AND_INTERVALS: a boolean that indicates whether or not this command this is supported by the real dome. I see two advantages:
* You can easily enable new commands as the vendor supports them.
* It reduces duplication.

python/lsst/ts/mtdome/mtdome_csc.py
```diff
@@ -304,7 +327,11 @@ def __init__(
         # with velocity == 0.0 since that may introduce large oscillations.
         self.current_moveAz_command = MoveAzCommandData()
 
-        self.log.info("DomeCsc constructed")
+        # Keep track of the commands that have been sent and that haven't been
+        # replied to yet.
```
r-owen, line 331: Please document what the keys are. I am guessing command id?

python/lsst/ts/mtdome/mtdome_csc.py
```diff
@@ -374,71 +401,88 @@ async def _set_maximum_motion_values(self) -> None:
...
+                host = tcpip.DEFAULT_LOCALHOST
                 port = self.mock_port
             else:
                 assert self.config is not None
+                host = self.config.host
                 port = self.config.port
-            self.mock_ctrl = MockMTDomeController(port)
+            self.mock_ctrl = MockMTDomeController(
+                host=host, port=port, log=self.log, name="MockMTDomeController"
+            )
```
r-owen, line 485: When you are starting your own mock controller here I very strongly suggest always using port 0 and the default host (which is likely tcpip.DEFAULT_LOCALHOST), because you don't want anybody other than the CSC talking to this internal mock controller.

I also strongly recommend eliminating self.mock_port entirely. If you want to run an *external* mock server (not managed by this CSC) then do the standard thing and set the config to point to it. That simplifies this code. You still need 2 simulation modes, but that is only to indicate which commands you can expect to be present. But it greatly simplifies the code to handle that mode.

python/lsst/ts/mtdome/mtdome_csc.py
```diff
@@ -511,48 +555,64 @@ async def write_then_read_reply(
...
+            command=command, tai=utils.current_tai()
+        )
+        command_dict = dict(commandId=command_id, command=command, parameters=params)
         async with self.communication_lock:
             try:
-                assert self.writer is not None
+                assert self.client is not None
             except AssertionError as e:
-                await self.fault(code=3, report=f"Error writing command {st}: {e}")
+                await self.fault(
```
r-owen, line 567: Instead of try/except consider something like this:
```
if self.client is None:
    await self.fault(...)
    raise RuntimeError("No client")
```

python/lsst/ts/mtdome/mtdome_csc.py
```diff
@@ -511,48 +555,64 @@ async def write_then_read_reply(
...
-                        self.reader.readuntil(b"\r\n"), timeout=_TIMEOUT
+                    await self.client.write_json(data=command_dict)
+                    data = await asyncio.wait_for(
+                        self.client.read_json(), timeout=_TIMEOUT
                     )
                 except (
                     asyncio.exceptions.IncompleteReadError,
                     asyncio.exceptions.TimeoutError,
-                    ConnectionResetError,
+                    ConnectionError,
```
r-owen, line 585: Do you need to be so specific, or can you just catch Exception? What exceptions are you trying allow to propagate up without going to fault?

python/lsst/ts/mtdome/mtdome_csc.py
```diff
@@ -511,48 +555,64 @@ async def write_then_read_reply(
...
+                match response:
+                    case ResponseCode.INCORRECT_PARAMETERS:
+                        error_message = f"{command=} has incorrect parameters."
+                    case ResponseCode.INCORRECT_SOURCE:
+                        error_message = f"{command=} was sent from an incorrect source."
+                    case ResponseCode.INCORRECT_STATE:
+                        error_message = f"{command=} was sent for an incorrect state."
+                    case _:
+                        error_message = f"{command=} is not supported."
+                raise ValueError(error_message)
```
r-owen, line 615: I gently suggest a case or dict of suffixes:
```
error_suffix = {
    ResponseCode.INCORRECT_PARAMETERS: "has incorrect parameters.",
    ResponseCode.INCORRECT_SOURCE:, "was sent from an incorrect source.",
    ...
}.get(respose, "is not supported.")
raise ValueError(f"{command=} {error_suffix"}"
```
Case works fine as well; the point is to reduce duplication of f"{command=}".

python/lsst/ts/mtdome/schema/lcs_status.py
```diff
@@ -19,18 +19,22 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
+__all__: list[str] = []
```
r-owen, line 22: Given that you have no symbols to import, does it really make sense to do `from lcs_status import *` in `__init__.py` (of this and most of the modules in `schema/`)? I worry it may confuse users, so I lean towards using `import lcs_status` instead. The down side is that doesn't do the right thing if in the future somebody adds symbols to one or more of these modules that you do want hoisted.

tests/test_mock_controller.py
```diff
@@ -43,85 +38,93 @@
...
-        pass
+
+class MockTestCase(tcpip.BaseOneClientServerTestCase):
+    server_class = mtdome.MockMTDomeController
 
     async def asyncSetUp(self) -> None:
-        self.ctrl = None
-        self.writer = None
-        port = 0
+        self.command_id: typing.Any = -1
```
r-owen, line 48: Note that there is nothing async in here. Perhaps setUp?

Please add a code comment explaining typing.Any (I think the point is to be able to specify invalid data types) and explain that valid data is an int.

ups/ts_mtdome.table
```diff
@@ -4,6 +4,8 @@
 #   the "sconsUtils" package.
 setupRequired(sconsUtils)
 setupRequired(ts_salobj)
+setupRequired(ts_tcpip)
```
r-owen, line 7: I gently suggest removing scons support while you are messing around in this package (in which case remove `setupRequired(sconsUtils)` and delete files `SConstruct` and `tests/SConscript`).

r-owen (approved):
I have a few concerns and am on the fence about approve vs request changes. Have a look and see what you think. I am happy to have another look.

doc/response_codes.rst
```diff
@@ -16,3 +16,5 @@ This list, as with the commands and statuses, is not necessarily complete and wi
     0, "OK", "Command received correctly and is being executed."
     2, "Unsupported command", "A command was sent that is not supported by the lower level component, for instance park is sent to LCS or 'mooveAz' instead of 'moveAz' to AMCS."
     3, "Incorrect parameter(s)", "The command that was sent is supported by the lower level component but the parameters for the command are incorrect. This can mean not enough parameters, too many parameters or one or more parameters with the wrong name."
+    4, "Incorrect source", "The current command source is not valid, e.g. a remote command arrives while the system is operated in local mode, like the push buttons for the Aperture Shutters."
+    5, "Incorrect state", "The current command cannot be executed in current state, e.g. moveAz when the AMCS is in fault state."
```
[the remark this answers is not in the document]
wvreeven: Great idea! Done.

python/lsst/ts/mtdome/mock_controller.py
```diff
@@ -182,117 +183,90 @@ async def start(self, keep_running: bool = False) -> None:
...
+                        # the command never is received or the reply never
+                        # sent.
+                        return
+
+                    func = self.dispatch_dict[cmd]
+                    kwargs = items["parameters"]
+                    if cmd.startswith("status"):
+                        # The status commands take care of sending a reply
+                        # themselves.
+                        send_response = False
```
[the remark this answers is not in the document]
wvreeven: Unfortunately using `return` here will omit executing the command.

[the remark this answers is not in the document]
wvreeven: Wait, I found a way to make sure that the command gets executed and that I can drop the `send_response` variable.

python/lsst/ts/mtdome/mock_controller.py
wvreeven: It turns out that the `host` constructor argument is mandatory because of the `create_server` method in `tcpip.BaseOneClientServerTestCase` so I kept it and set it by default to `tcpip.DEFAULT_HOST` but I eliminated the constructor argument everywhere where a `MockMTDomeController` is constructed.

python/lsst/ts/mtdome/mock_llc/mock_motion/base_llc_motion_with_crawl.py
wvreeven: I added because PyCharm complains if it isn't there. But I removed it again because it, indeed, is inherited.

python/lsst/ts/mtdome/mtdome_csc.py
wvreeven: No need at all. I reduced it to `Exception`.

python/lsst/ts/mtdome/mtdome_csc.py
wvreeven: Turns out that there is exactly one use case for `self.mock_port`: when the mock controller is started by the unit test and the CSC needs to connect to that. This is useful for testing disconnections. In the end I did eliminate `self.mock_port` and am simply loading the test config file to get the port from there so it can be used by the mock controller.

python/lsst/ts/mtdome/schema/lcs_status.py
wvreeven: It turns out to be even simpler: the `__all__` can be removed from all schema files and needs to stay in `registry.py`. Then all `from .xxxxx import *` can be removed, except again for the registry file. The schemas are only supposed to be available via the schema registry and this actually makes that stronger.

tests/test_mock_controller.py
wvreeven: The `MockTestCase` inherits from `tcpip.BaseOneClientServerTestCase` which certainly is an async test case, so `asyncSetUp` needs to stay.

I added a comment to `self.command_id`.

wvreeven:
I have addressed all your comments and would appreciate another look.

python/lsst/ts/mtdome/mock_controller.py
r-owen: Good point, but: I actually fixed it as part of ts_tcpip 1.1.1 -- though as a commit that I had not pushed when you reviewed it (oops). It's a really small change: if the server class is OneClientServer and create_server's `**kwargs` does not include `host` then add it. So usually there will be no host argument.

I discovered it in ts_atdome, and when I fixed it I did not need to update ts_atdome for ts_tcpip 1.1.1.

Feel free to leave the host argument; I just wanted you to know that it is not actually required any more.

python/lsst/ts/mtdome/mock_llc/mock_motion/base_llc_motion_with_crawl.py
r-owen: Thank you for the explanation. Sorry about PyCharm.

python/lsst/ts/mtdome/mtdome_csc.py
r-owen: I realized later (too late to want to log back in) that it might be nicer to have a constant that is a set or frozen set of unsupported commands. Then you can iterate over METHODS_AND_INTERVALS_FOR_COMMISSIONING while excluding anything in the unsupported command set (which will shrink to nothing over time as the vendor adds support for commands).

Still, adding a flag is a quite reasonable solution. Just passing it along in case you prefer it enough to redo it again.

python/lsst/ts/mtdome/mtdome_csc.py
```diff
@@ -374,71 +394,85 @@ async def _set_maximum_motion_values(self) -> None:
...
+        interval: `float`
+            The interval (sec) at which to run the status method.
+
+        """
+        try:
+            while True:
+                await method()
+                await asyncio.sleep(interval)
+        except Exception:
+            self.log.exception(f"one_periodic_task({method}) failed.")
```
r-owen, line 443: Please state in the log message that you are continuing (despite the error), as otherwise I think readers of the message will expect that the task died at this point.

Also I gently suggest including repr(error) in the message (e.g. as `{e!r}`). That way it shows up in LOVE log message displays and simple Chronograf dashboards.

python/lsst/ts/mtdome/schema/lcs_status.py
r-owen: That sounds great, but...how do these modules get loaded so that they can register their contents? I assumed that was why `__init__.py` was importing them.

tests/test_mock_controller.py
r-owen: Actually `setUp` and `tearDown` work fine in async unit tests. unittest.IsolatedAsyncioTestCase (spelling?) *adds* `asyncSetUp` and `asyncTearDown`, but also runs the standard synchronous `setUp` and `tearDown`.

I have found one situation where I needed `asyncSetUp` when it's not obvious: constructing an object that needs a running event loop in the constructor. But I don't think this is such a case.

r-owen (approved):
Looks really nice. Thanks for all the improvements to already really nice code.

python/lsst/ts/mtdome/schema/lcs_status.py
wvreeven: They register themselves. Then the CSC and MockController code import the registry and they have all schemas at their disposal. You taught me that trick :)

tests/test_mock_controller.py
wvreeven: That is really good to know!

python/lsst/ts/mtdome/mtdome_csc.py
wvreeven: The exception is caught outside the loop, so the task does not continue. I've made that clear and also added repr(error) in the statement as you suggested.

python/lsst/ts/mtdome/mock_controller.py
wvreeven: I indeed missed that. I've removed the host argument now.

python/lsst/ts/mtdome/schema/lcs_status.py
r-owen: I am still confused. I don't understand how they can register themselves without somebody importing these modules. (It's not meant to block merging; I am just asking for my own curiosity).

python/lsst/ts/mtdome/mtdome_csc.py
r-owen: Oops. Of course! No need to clarify anything; I should have seen that. Sorry!

python/lsst/ts/mtdome/schema/lcs_status.py
wvreeven: See for instance these lines:

https://github.com/lsst-ts/ts_mtdome/blob/develop/python/lsst/ts/mtdome/schema/amcs_status.py#L24-L26

No need for `__all__` in that file because nothing is exported. The code simply adds the contents of the schema file to the registry and the registry does get exported.

PR #105, opened 2023-10-31
Tickets/dm 40861

wvreeven:


python/lsst/ts/mtdome/power_management/power_management_handler.py
```diff
@@ -0,0 +1,486 @@
...
+                )
+            case PowerManagementMode.EMERGENCY:
+                scheduled_command = await self.get_next_command_emergency(
+                    current_power_draw
+                )
+            case PowerManagementMode.NO_POWER_MANAGEMENT | _:
+                _, scheduled_command = await self.command_queue.get()
+        return scheduled_command
+
+    async def get_next_command_operations(
```
tribeiro, line 201: If I understood this correctly, this method is not only "getting" the next command but also "issuing" the command. If this is correct, I think it might be a bit misleading to call this method "get_<something>". Maybe call it `execute_next_command_operations`?

python/lsst/ts/mtdome/power_management/power_management_handler.py
```diff
@@ -0,0 +1,486 @@
...
+                    return await self.issue_scheduled_command_or_stop_commands(
+                        scheduled_command,
+                        current_power_draw,
+                        [],
+                        [LlcName.APSCS, LlcName.LWSCS, LlcName.LCS],
+                    )
+            case _:
+                return scheduled_command
+
+    async def get_next_command_maintenance(
```
tribeiro, line 277: See my comment above about using "get_<something>" on the name of this method.

python/lsst/ts/mtdome/power_management/power_management_handler.py
```diff
@@ -0,0 +1,486 @@
...
+                | CommandName.SET_LOUVERS
+                | CommandName.CLOSE_LOUVERS
+            ):
+                # These commands are not allowed and therefore are ignored. By
+                # doing so, they automatically are removed from the queue.
+                return None
+            case _:
+                return scheduled_command
+
+    async def get_next_command_emergency(
```
tribeiro, line 351: See my comment above about using "get_<something>" to name this method.

tribeiro (approved):
Looks good. I left a few inline comments I hope you will consider before merging.

bin/run_mtdome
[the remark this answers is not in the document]
wvreeven: Yes, I use it to run the MTDome CSC locally when testing it with the EIE LabVIEW simulator.

doc/power_management.rst
```diff
@@ -0,0 +1,262 @@
...
+Introduction
+============
+
+Power for the rotating part of the dome is provided via a slip ring with limited capacity.
+The slip ring has a limited power capacity of 78 kW continuous, and 100 kW Peak for 6 minutes.
+Since the slip ring capacity has not been derated for the thinner air at altitude, this rating is should not be considered conservative.
+The total power demand of all the components supplied by the slip ring is greater than the capacity of the slip ring.
+Since this total demand of 177.85 kW is significantly greater than the slip rings capacity, power management to limit the simultaneous operation of the systems is required.
+
+The slip ring demand is dominated by three systems, the Light Wind Screen (LWS), the Louvers, and the Fans.
```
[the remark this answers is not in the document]
wvreeven: Good idea. I used a colon `:` instead.

python/lsst/ts/mtdome/power_management/power_management_handler.py
```diff
@@ -0,0 +1,486 @@
...
+        Returns
+        -------
+        ScheduledCommand | None
+            The next command to be issued, or None if no commands currently can
+            be executed or are scheduled.
+        """
+        if self.command_queue.empty():
+            return None
+
+        match self.power_management_mode:
```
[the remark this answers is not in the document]
wvreeven: Good suggestion. I originally had a different approach in mind and forgot to consolidate this.

python/lsst/ts/mtdome/power_management/power_management_handler.py
wvreeven: Good point but it actually doesn't issue the command. That is done by the cRIO. So `get_` is correct for this and the other two PowerManagementMode-specific methods. I have renamed the `issue_scheduled_command_or_stop_commands` to `generic_get_scheduled_command_or_stop_commands` to reflect that there too.

python/lsst/ts/mtdome/power_management/power_management_handler.py
wvreeven: See my reply above.

python/lsst/ts/mtdome/power_management/power_management_handler.py
wvreeven: See my reply above.

python/lsst/ts/mtdome/power_management/power_management_handler.py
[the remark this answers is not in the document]
wvreeven: After making the change I actually realized that the options are NOT doing the same. Each option has a specific `get_next_command_<PowerManagementMode>` command which do different things. So I consolidated the three cases as follows:
```
            case (
                PowerManagementMode.OPERATIONS
                | PowerManagementMode.MAINTENANCE
                | PowerManagementMode.EMERGENCY
            ):
                func = getattr(
                    self, f"get_next_command_{self.power_management_mode.name.lower()}"
                )
                scheduled_command = await func(current_power_draw)
```

PR #106, opened 2023-12-01
Tickets/dm 41563

wvreeven:


python/lsst/ts/mtdome/power_management/slip_ring.py
```diff
@@ -0,0 +1,246 @@
...
+
+class SlipRing:
+    """A state machine implementation representing a slip ring.
+
+    The slip ring of the MTDome can handle continuous power up to a limit of 78
+    kW and temporary power up to 100 kW for up to 6 minutes. After going over
+    the continuous limit, it needs to cool down for up 4 minutes.
+
+    The exact cooling down model is unknown so a model proportional to the
+    maximum drawn power and the amlunt of time over the continuous limit is
```
tribeiro, line 46: typo "amlunt" -> "amount"?

python/lsst/ts/mtdome/power_management/slip_ring.py
```diff
@@ -0,0 +1,246 @@
...
+
+        A simple model is assumed to calculate the cool down time. This needs
+        to be replaced with the real cool down model.
+
+        Returns
+        -------
+        float
+            The cool down time for the slip ring.
+        """
+        if self.max_power_drawn < CONTINUOUS_SLIP_RING_POWER_CAPACITY:
```
tribeiro, line 111: Consider replace this if/else with something like:

```
power_over_limit_fraction = 0.0 if self.max_power_drawn < CONTINUOUS_SLIP_RING_POWER_CAPACITY else (
                self.max_power_drawn - CONTINUOUS_SLIP_RING_POWER_CAPACITY
            ) / OVER_LIMIT_POWER_AVAILABLE
```

This is a bit more expressive.

python/lsst/ts/mtdome/power_management/slip_ring.py
```diff
@@ -0,0 +1,246 @@
...
+            return await self._handle_disabled()
+        match self.state:
+            case SlipRingState.BELOW_LOW_LIMIT:
+                return await self._handle_below_low_mimit(current_power_draw)
+            case SlipRingState.OVER_LOW_LIMIT:
+                return await self._handle_over_low_mimit(current_power_draw)
+            case SlipRingState.COOLING_DOWN:
+                return await self._handle_cooling_down()
+            case _:
+                raise ValueError(
```
tribeiro, line 152: Document exception. 

Also, I'd return a `RuntimeError` instead of `ValueError`.

python/lsst/ts/mtdome/power_management/slip_ring.py
```diff
@@ -0,0 +1,246 @@
...
+                    f"encountered unknown SlipRingState {self.state.name}."
```
tribeiro, line 153: I think you risk having a second exception here if `self.state` is not of `SlipRingState` type. I know you are using mypy to check but better safe than sorry. Consider replacing `{self.state.name}` with `{self.state!r}`, this will give you almost the same behavior if this is a `SlipRingState` object and will be safe in case it is not.

python/lsst/ts/mtdome/power_management/slip_ring.py
```diff
@@ -0,0 +1,246 @@
...
+            # Keep track of the maximum drawn power for calculating the cool
+            # down time.
+            self.max_power_drawn = max(self.max_power_drawn, current_power_draw)
+            # When the power draw started exceeding the continuous limit?
+            self.over_limit_start_time = utils.current_tai()
+            self.state = SlipRingState.OVER_LOW_LIMIT
+
+        return MAXIMUM_SLIP_RING_POWER_CAPACITY
+
+    async def _handle_over_low_mimit(self, current_power_draw: float) -> float:
```
tribeiro, line 192: typo in function name? `mimit` -> `limit`?

python/lsst/ts/mtdome/power_management/slip_ring.py
```diff
@@ -0,0 +1,246 @@
...
+        else:
+            # Keep track of the maximum drawn power for calculating the cool
+            # down time.
+            self.max_power_drawn = max(self.max_power_drawn, current_power_draw)
+            return MAXIMUM_SLIP_RING_POWER_CAPACITY
+
+    async def _handle_cooling_down(self) -> float:
+        """Handle the COOLING_DOWN case.
+
+        Transit to BELOW_LOW_LIMIT if not cooling anymore.
```
tribeiro, line 232: Transit -> Transition ?

python/lsst/ts/mtdome/power_management/slip_ring.py
```diff
@@ -0,0 +1,246 @@
...
+        self.state = SlipRingState.BELOW_LOW_LIMIT
+
+        self.below_limit_start_time = 0.0
+        self.over_limit_start_time = 0.0
+        self.time_over_limit = 0.0
+        self.cooling_down_time_needed = 0.0
+
+        self.max_power_drawn = 0.0
+
+    async def get_cool_down_time(self) -> float:
```
tribeiro, line 100: Is there a particular reason these methods are all coroutines? Doesn't look like any of them need to be. Are you trying to future proof this against potentially having to wait for data to compute these quantities and/or do state transitions?

tribeiro (approved):
Looks good, but please, take a look at my inline comments before merging.

python/lsst/ts/mtdome/power_management/slip_ring.py
wvreeven: Good catch, thanks!

python/lsst/ts/mtdome/power_management/slip_ring.py
wvreeven: No reason at all. I made them into normal methods instead.

PR #118, opened 2024-04-10
Add do_setPowerManagementMode command.

wvreeven:


tribeiro (commented):
We just released new xml to the summit. I understand the desire to update the interface but please, make sure this is still compatible with the current released xml. Otherwise we won't be able to make any patch to the summit systems.

python/lsst/ts/mtdome/mtdome_csc.py
```diff
@@ -1184,6 +1214,46 @@ async def do_inflate(self, data: salobj.BaseMsgType) -> None:
...
+        self.assert_enabled()
+
+        new_mode = PowerManagementMode(data.mode)
+
+        if new_mode == PowerManagementMode.NO_POWER_MANAGEMENT:
+            raise salobj.ExpectedError(
+                "Will not set PowerManagementMode to NO_POWER_MANAGEMENT."
+            )
+
+        if new_mode == self.power_management_mode:
```
tribeiro, line 1245: Consider using an elif instead.

tribeiro (approved):
Looks great! thanks for the backward compatibility work! I just have one small suggestion to replace an if with elif.. feel free to ignore it if you want.

PR #127, opened 2024-12-04
Ensure that aperture shutter position values cannot be negative.

wvreeven:
I implemented a jitter around the position of the aperture shutter. Then I implemented code that rounds off telemetry values to a certain amount of decimals and made sure that any `-0.0` values will be `0.0`.

### DIFF
--- a/doc/version_history.rst
+++ b/doc/version_history.rst
@@ -7,6 +7,19 @@ Version History
 ###############
 
 =======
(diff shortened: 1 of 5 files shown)


sebastian-aranda (approved):
Thanks for working on this @wvreeven! PR looks good to me :+1:.

PR #131, opened 2025-02-11
Tickets/dm 48703

wvreeven:


python/lsst/ts/mtdome/mtdome_csc.py
```diff
@@ -650,84 +577,208 @@ async def do_setPowerManagementMode(self, data: salobj.BaseMsgType) -> None:
...
-        level component.
+        Parameters
+        ----------
+        status : `dict`[`str`, `typing.Any`]
+            The status.
         """
-        await self.request_and_send_llc_status(LlcName.AMCS, self.tel_azimuth)
+        if "exception" in status:
+            self.log.error(status["exception"])
+            await self.go_fault(self.status_amcs)
```
teweitsai, line 590: You could abstract this part for the other functions to reuse.

teweitsai (approved):
Looks good. Do you want to update the schema for me to use the `gui_port` in ts_config_mttcs?

wvreeven:
It is not `gui_port` but `eui_port`. And yes, please use that.

python/lsst/ts/mtdome/mtdome_csc.py
wvreeven: I could but I am lazy ;)

PR #138, opened 2025-04-16
Tickets/dm 50222

wvreeven:


teweitsai (approved):
Looks good. I understand why we would like to have this fix. But I think we might need to have some balance between the interruption and the possibility that there is some issue at the moment and we do need to fault the system to avoid the further possible damage. For the system such as the AMCS, I would prefer to just fail the CSC if we lost the status and could not get the new one in a timeout period instead.

wvreeven:
I understand what you indicate about the balance and about going to FAULT. The challenge is to be able to keep on using other MTDome subsystems (like aperture shutter, louvers and light/.wind screen) even when azimuth rotation is failing. Telemetry is incredibly important but being able to be on sky as well. We probably need to discuss this with Tiago.

PR #140, opened 2025-06-05
Ensure that aperture shutter InPosition is correct.

wvreeven:


mareuter (approved):
Looks good to me. BTW, news creation job failed. :-)

PR #182, opened 2026-03-09
Tickets/osw 1927

wvreeven:


python/lsst/ts/mtdome/mtdome_csc.py
```diff
@@ -112,7 +115,15 @@ def __init__(
...
-        setattr(self, "do_resetDrivesLouvers", self._do_resetDrivesLouvers)
+        component_info = ComponentInfo(name="MTDome", topic_subname=None)
+
+        # TODO OSW-1949 Remove backward compatibility with XML 26.0.
+        if "cmd_resetDrivesEl" in component_info.topics:
+            setattr(self, "do_resetDrivesEl", self._do_resetDrivesEl)
+
+        # TODO OSW-1949 Remove backward compatibility with XML 26.0.
+        if "cmd_calibrateEl" in component_info.topics:
+            setattr(self, "do_calibrateEl", self._do_calibrateEl)
```
teweitsai: I think you can use the `extra_commands` argument in `ConfigurableCsc.__init__()` instead:
https://github.com/lsst-ts/ts_salobj/blob/develop/python/lsst/ts/salobj/configurable_csc.py#L76-L78

python/lsst/ts/mtdome/mtdome_csc.py
```diff
@@ -506,9 +518,26 @@ async def do_resetDrivesAz(self, data: salobj.BaseMsgType) -> None:
         assert self.mtdome_com is not None
         await self.call_method(method=self.mtdome_com.reset_drives_az, reset=reset_ints)
 
+    # TODO OSW-1949 Remove backward compatibility with XML 26.0.
+    async def _do_resetDrivesEl(self, data: salobj.BaseMsgType) -> None:
```
teweitsai: If you use the `extra_commands` above, you can do `async def do_resetDrivesEl()` here.

teweitsai (approved):
Looks good.

python/lsst/ts/mtdome/mtdome_csc.py
wvreeven: Thanks, that is very useful!

python/lsst/ts/mtdome/mtdome_csc.py
wvreeven: Thanks, that is very useful!

PR #187, opened 2026-07-19
Tickets/osw 2678

wvreeven:


python/lsst/ts/mtdome/mtdome_csc.py
```diff
@@ -58,6 +57,9 @@
 # Time [s] to sleep after an exception in a status command.
 EXCEPTION_SLEEP_TIME = 1.0
 
+# A small but larger than zero threshold taking jitter into account.
+SHUTTER_POSITION_THRESHOLD = -0.1
```
teweitsai: Just want to make sure the threshold is -0.1 because the comment says: "larger than zero."

python/lsst/ts/mtdome/mtdome_csc.py
```diff
@@ -823,20 +821,84 @@ async def send_llc_status_telemetry_and_events(
...
+                MotionState.STOPPED.name,
+                MotionState.ENGAGING_BRAKES.name,
+                MotionState.BRAKES_ENGAGED.name,
+                MotionState.GO_STATIONARY.name,
+                MotionState.DISABLING_MOTOR_POWER.name,
+                MotionState.MOTOR_POWER_OFF.name,
+                MotionState.OPEN.name,
+            ]:
+                if status["positionActual"][i] > SHUTTER_POSITION_THRESHOLD:
+                    self.last_seen_shutter_positions[i] = status["positionActual"][i]
```
teweitsai, line 877: If the CSC wants to track the previous reported shutter positions, I would prefer to have the `SHUTTER_POSITION_THRESHOLD` to be `0.1` instead of `-0.1`.

In addition, maybe consider:

```python
if SHUTTER_POSITION_THRESHOLD <= status["positionActual"][i] <= (100 - SHUTTER_POSITION_THRESHOLD):
    self.last_seen_shutter_positions[i] = status["positionActual"][i]
```

python/lsst/ts/mtdome/mtdome_csc.py
```diff
@@ -823,20 +821,84 @@ async def send_llc_status_telemetry_and_events(
...
+                    MotionState.MOTOR_POWER_ON.name,
+                    MotionState.GO_NORMAL.name,
+                    MotionState.DISENGAGING_BRAKES.name,
+                    MotionState.BRAKES_DISENGAGED.name,
+                    MotionState.CLOSING.name,
+                ]
+                and status["positionActual"][i] == mtdomecom.APSCS_CLOSED_POSITION
+            ) or status["positionActual"][i] < SHUTTER_POSITION_THRESHOLD < self.last_seen_shutter_positions[
+                i
+            ]:
```
teweitsai: I am confused with this line:

```python
status["positionActual"][i] < SHUTTER_POSITION_THRESHOLD < self.last_seen_shutter_positions[i]
```

If you set `SHUTTER_POSITION_THRESHOLD = -0.1`, I felt you could do the following only since the `self.last_seen_shutter_positions[i]` should be always > `SHUTTER_POSITION_THRESHOLD`:

```python
status["positionActual"][i] < SHUTTER_POSITION_THRESHOLD
```

python/lsst/ts/mtdome/mtdome_csc.py
```diff
@@ -823,20 +821,84 @@ async def send_llc_status_telemetry_and_events(
...
+                # The positionActual here is negative, so we need to subtract
+                # the absolute value.
+                status["positionActual"][i] = self.last_seen_shutter_positions[i] - math.fabs(
+                    status["positionActual"][i]
+                )
```
teweitsai, line 895: I am confused here because if the issue is to resolve: "This will lead to the shutter positions resetting to zero while they should be at the same position as before.", I would assume the code here is to report the previous cached value to be:

```python
status["positionActual"][i] = self.last_seen_shutter_positions[i]
```

Or maybe I confused with myself for the intention of code here because the relative encoder is used in the shutter. But if this is the issue, I felt this should be the controller to resolve this...

Maybe you could explain more here.

teweitsai (approved):
Please help to explain a litter more for the consideration here. My understanding for this PR is that the CSC will cache the received shutter position value if it is larger than the jitter. If the power of shutter is lost, the cRIO controller reports the value to be 0 and the CSC will use the previous cached value instead to report to EFD. But I am not sure why the use of cached value will need to subtract the received value. In addition, I felt if the cRIO can still report to the CSC, it should report the cached shutter position value to the CSC instead of letting the CSC to do so. Thanks!

python/lsst/ts/mtdome/mtdome_csc.py
wvreeven: I updated the comment.

python/lsst/ts/mtdome/mtdome_csc.py
wvreeven: I tried this at first and then the code doesn't behave well when 0 <= position <= threshold. I only could get it to work reliably well when taking a small but negative threshold.

python/lsst/ts/mtdome/mtdome_csc.py
wvreeven: Agreed, thanks! I updated that line.

python/lsst/ts/mtdome/mtdome_csc.py
wvreeven: Perhaps the name `last_seen_shutter_position` is poorly chosen. What the cRIO currently reports for the shutter door positions is 0 when the power is lost and a negative number when the doors are closing. If the door were fully opened (last seen position == 100) then the positions will be -100 when fully closed. If the door were partially open (for instance last seen position == 63) then the position will be -63 when fully closed. However, the positionActual that _**should be**_ in the telemetry is 0 so that is `positionActual - last_seen_position`.

teweitsai: If the controller reports to have the value of -100 or -63, I think the code of controller has the issue. Maybe we should discuss this with Lorenzo since only the value of 0-100 with some jitters makes sense here instead of such big negative values.

wvreeven: You are right that the controller has the problem. And we did discuss this with Lorenzo. It is easier and safer to fix this in the CSC than in the cRIO which is why I am implementing this now. The code, however, probably should be in MTDomeComm instead of the CSC. That way the GUI benefits from this as well. Sorry, I didn't think of that. I'll open a new ticket.

PR #191, opened 2026-08-04
Tickets/osw 2706

wvreeven:


python/lsst/ts/mtdome/mtdome_csc.py
```diff
@@ -773,7 +800,14 @@ async def status_moncs(self, status: dict[str, typing.Any]) -> None:
...
             await self.log_status_exception(status)
-        await self.send_llc_status_telemetry_and_events(LlcName.MONCS, status, self.tel_interlocks)
+
+        # TODO OSW-2451 Remove backward compatibility.
+        if hasattr(self, "evt_interlocksAMCS"):
+            for evt_name in MONCS_EVENT_NAMES.keys():
+                telemetry = status[evt_name]
+                evt = getattr(self, f"evt_{evt_name}")
+                evt_sent = await evt.set_write(**telemetry)
+                self.log.debug(f"{evt_name}: {evt_sent}")
```
teweitsai: I have the performance concern here. There are around 30 events in `MONCS_EVENT_NAMES`. It might be too much to do this continuously. In addition, the search of `getattr()` in Python is pretty slow.

Please consider to check the performance (maybe try the base teststand) to see the response speed of CSC command that it could be immediately or not on the base teststand. If there is the performance impact, please consider to use the bit value instead of differentiating each interlock condition as a single event.

Thanks!

teweitsai (approved):
Looks good.

python/lsst/ts/mtdome/mtdome_csc.py
wvreeven: Please note that these are events. The values of the items will not change often (perhaps a few times per day) and salobj makes sure to only send the event to Kafka if one or more item values change. So in general this will be fast and only occasionally an event is sent to Kafka. Good point about `getattr`. I will modify that and will explicitly send each event instead of using a loop.

